PR #11, opened 2021-03-15
Fix ref counting bug with Zeek 3.0.

rsmmr:
I believe this is what lets https://github.com/zeek/spicy-analyzers/pull/12/checks?check_run_id=2101642810 fail. We'll see ...

### DIFF
--- a/plugin/include/zeek-compat.h
+++ b/plugin/include/zeek-compat.h
@@ -310,7 +310,7 @@ inline auto EnumVal_GetType(::zeek::EnumVal* v) { return v->Type(); }
 inline auto EventHandler_GetType(::zeek::EventHandlerPtr ev, bool check_export = true) {
     return ev->FType(check_export);
 }
-inline auto FileAnalysisComponentTag_AsVal(const ::zeek::file_analysis::Tag& t) { return t.AsEnumVal(); }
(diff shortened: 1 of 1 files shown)


plugin/include/zeek-compat.h
```diff
@@ -310,7 +310,7 @@ inline auto EnumVal_GetType(::zeek::EnumVal* v) { return v->Type(); }
 inline auto EventHandler_GetType(::zeek::EventHandlerPtr ev, bool check_export = true) {
     return ev->FType(check_export);
 }
-inline auto FileAnalysisComponentTag_AsVal(const ::zeek::file_analysis::Tag& t) { return t.AsEnumVal(); }
+inline auto FileAnalysisComponentTag_AsVal(const ::zeek::file_analysis::Tag& t) { return t.AsEnumVal()->Ref(); }
```
bbannier, line 313: This does indeed seem to fix the issue (assuming the issue wasn't flaky).

Since most of the wrappers here return `auto` it isn't very clear what type we return, and the returned type might even change between Zeek versions. Maybe we should introduce hardcoded types here to introduce a firewall against Zeek API changes.

bbannier (approved):

plugin/include/zeek-compat.h
rsmmr: It actually does change between Zeek versions already (raw pointers to smarts ptrs). A bunch of this will go away eventually once we no longer need to support 3.x. so don't think it's worth bothering.

PR #12, opened 2021-03-15
Split CI zkg run into separate test and install steps.

rsmmr:
Installing with --force lets it ignore errors.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -15,8 +15,11 @@ package_ci_task:
     cpu: 2
     memory: 12G
 
+  test_script:
+    - zkg test .
+
   install_script:
-    - zkg install --force .
+    - zkg install --force --skiptests .
 
   check_script:
     - zeek -N _Zeek::Spicy
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -25,6 +25,9 @@ package_ci_task:
     - zeek -N _Zeek::Spicy
     - zeek local
 
+  always:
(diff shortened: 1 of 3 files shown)


.cirrus.yml
```diff
@@ -15,13 +15,19 @@ package_ci_task:
     cpu: 2
     memory: 12G
 
+  test_script:
+    - zkg test .
+
   install_script:
-    - zkg install --force .
+    - zkg install --force --skiptests .
```
bbannier, line 22: I am confused now why we install with `--force` if it in turn forces us to deviate from the installation method a user would default to (ideally we'd mirror what a user would do during installation as much as possible and only add our own steps _in addition to that_).

Could you add a comment somewhere here capturing the reasoning?

bbannier (approved):

.cirrus.yml
rsmmr: "Force" avoids interactive prompts ("do you want to install package X?").

bbannier: Hmm, that's unfortunate. Could we maybe instead work around that with e.g.,

    $ yes | zkg install .
    $ echo Y | zkg install .

I think I tried using `yes` previously when this CI also had other issues, so maybe we need to work around issues with `echo Y`.

Not something super critical which needs to block this PR, but IMO still nice to have.

rsmmr: not too tempted to try that :) I think the force is fine

PR #17, opened 2021-03-17
Silence clang-tidy.

rsmmr:
I don't see why it flags this, but it doesn't matter anyways as its a
one-time allocation, of which we have more leaking in Zeek.

### DIFF
--- a/plugin/src/plugin.cc
+++ b/plugin/src/plugin.cc
@@ -476,7 +476,10 @@ void plugin::Zeek_Spicy::Plugin::InitPostScript() {
     // type. Give it a dummy event type in that case, so that we don't walk
     // around with a nullptr.
     for ( const auto& [name, id] : _events ) {
-        if ( ! compat::ID_GetType(id) )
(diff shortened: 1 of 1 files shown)


plugin/src/plugin.cc
```diff
@@ -476,7 +476,10 @@ void plugin::Zeek_Spicy::Plugin::InitPostScript() {
     // type. Give it a dummy event type in that case, so that we don't walk
     // around with a nullptr.
     for ( const auto& [name, id] : _events ) {
-        if ( ! compat::ID_GetType(id) )
+        if ( ! compat::ID_GetType(id) ) //NOLINT(clang-analyzer-cplusplus.NewDeleteLeaks)
+            // clang-tidy reports a leak here even with >= 3.2. Not sure why because
+            // the memory should be managed. Either way, doesn't matter so we
+            // ignore.
```
bbannier, line 482: The overload we are using here should already be using a managed pointer, right? It might be that clang-tidy is unable to detect here that this custom smart pointer doesn't leak when used like we do here.

bbannier (approved):

plugin/src/plugin.cc
rsmmr: right indeed.

PR #21, opened 2021-03-18
Install Spicy from upstream package.

bbannier:


ci/Dockerfile
```diff
@@ -3,17 +3,12 @@
 
 FROM ubuntu:focal
 
-# Increase this to force a rebuild of the Docker image (in Cirrus, in particular).
```
rsmmr, line 6: I'd leave this in. We won't need it for Spicy anymore but could still want to update dependencies.

rsmmr (approved):
Cool, this looks excactly like I was hoping it would! :-)

PR #23, opened 2021-03-22
Add evt syntax for port ranges.

### LINKED ISSUE #22: support port range rather than just a single port
Protocols with which I am familiar, such as BACnet in its clause J.1.2, specify a port range rather than a single port. `B/IP devices shall also support a configurable UDP port number and shall support, at a minimum, values in the ranges 47808 - 47823 and 49152 - 65535. For B/IP devices that support multiple B/IP ports, the UDP port number for each B/IP port shall be settable across the above noted valid range.` 

And for Voice-over-IP, the Vonage FAQ states `Your network must allow inbound and outbound traffic from RTP ports 10,000 - 30,000 to prevent one-way or no audio on phone calls.` https://businesssupport.vonage.com/articles/answer/Port-Forwarding-807

Support for port range is here proposed, per https://github.com/zeek/spicy/pull/547#issuecomment-714271817, as Feature Request.

bbannier:
This patch adds support for specifying port ranges in evt files.

Closes #22.


bbannier:
The corresponding docs update is zeek/spicy#852.

tests/zeek/udp.zeek
```diff
@@ -17,7 +17,7 @@ public type Message = unit {
 # @TEST-START-FILE udp-test.evt
 protocol analyzer spicy::UDP_TEST over UDP:
     parse with UDPTest::Message,
-    port 31337/udp;
+    port 31337/udp-31340/udp;
```
rsmmr, line 20: That syntax with the two protocols feels a little bit odd, we could instead do `31337-31340/udp`. But I see from the code that parsing the current syntax is easier, and it's a rare use case to begin with; so fine leaving as is, just wanted to note.

rsmmr (dismissed):

tests/zeek/udp.zeek
rsmmr: Actually, just reading the doc PR: should we move the range parsing into the plural case: `ports 31337/udp-31340/udp`?

bbannier: The way I had refactored this already supported specifying either a port or a range, in both `port` or `ports` instructions. I have clarified the docs in zeek/spicy#852.

bbannier:
Could you make another pass @rsmmr? Since your last review I changed port ranges from half-closed to closed intervals so a range can span the whole domain of ports.

tests/zeek/udp.zeek
```diff
@@ -17,7 +17,8 @@ public type Message = unit {
 # @TEST-START-FILE udp-test.evt
 protocol analyzer spicy::UDP_TEST over UDP:
     parse with UDPTest::Message,
-    port 31337/udp;
+    port 11337/udp-11340/udp,
+    ports {31337/udp-31340/udp};
```
rsmmr, line 21: I'm confused about syntax now: are these two lines doing the same?

compiler/glue-compiler.cc
```diff
@@ -242,6 +242,35 @@ static hilti::rt::Port extract_port(const std::string& chunk, size_t* i) {
...
+    }
+
+    std::vector<hilti::rt::Port> result;
+
+    // Port ranges are a closed interval.
+    for ( auto port = start.port(); ! end || port <= end->port(); ++port ) {
+        result.emplace_back(port, start.protocol());
+        if ( ! end )
+            break;
+    }
```
rsmmr, line 269: Not following the termination condition here: is `! end || port <= end->port()` correct? And we have an `if` terminating in the `!end`, so won't get into 2nd round with that anyways.

tests/zeek/udp.zeek
bbannier: Yes, they are both supported (but they construct different ranges here).

On the implementation side this is possible by using identical code to extract both a single and multiple ports (here: `extract_ports` which always returns a vector of `Port`).

compiler/glue-compiler.cc
bbannier: I wanted to reuse the same loop to handle both the case of `end` being unset (single port) and being set (port range given). Unless I have made a mistake, the logic should be:

-  for `! end` we should make exactly one iteration of the loop which adds `start` to the result,
- else `end` will be set and we want to iterate until and including when `port == end.port()` (closed range).

The alternative here would be to place `start` in `result` before the loop, and then iterate from `start.port() + 1`. This would need additional handling for the case where `start.port() == 65535`, so I went for this logic.

If you think it would help to make this more explicit I can expand the code and/or add additional comments.

rsmmr: Ok, got it now.

tests/zeek/udp.zeek
rsmmr: Ok, the syntax remains a bit non-intuitive to me, but with the docs it's fine.

rsmmr (approved):

PR #29, opened 2021-04-06
Request less memory for Cirrus CI tasks.

bbannier:
Since recently Cirrus CI seems to enforce that not more than 4GB of
memory can be requested per cpu. We were requesting more than that for a
long time.

This patch brings the memory requested per cpi in line with what we do
in zeek/spicy, namely we now request 3GB/cpu, which means that we now
request 6GB of memory total matching the cpu=2 request.

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -3,11 +3,11 @@ all: test
 test: test-build
 
 test-build:
-	@btest -j -d
+	@btest -d
 
 # Might need to have SPICY_INSTALLATION_DIRECTORY set.
 test-install:
-	@btest -j -d -a installation
+	@btest -d -a installation
 
 clean:
 	@rm -f $(DIAG) .btest.failed.dat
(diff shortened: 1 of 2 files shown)


tests/Makefile
```diff
@@ -3,11 +3,11 @@ all: test
 test: test-build
 
 test-build:
-	@btest -j -d
+	@btest -d
```
rsmmr, line 6: I suppose this is necessary even thought we already have `HILTI_JIT_SEQUENTIAL` set?

rsmmr (approved):

tests/Makefile
bbannier: Yes, unfortunately it looks like it. When I first tried only setting the memory, the _standalone_ci_ task was still OOM'ing. I am not sure why this seems to be less of an issue for zeek/spicy's CI (maybe it has to do with us using more cpus there).

PR #31, opened 2021-04-26
 Provide convenience wrapper unit for passing content into Zeek's file analysis.

rsmmr:
This in on top of #30, and requires  https://github.com/zeek/spicy/pull/896 to be merged to work.


plugin/include/cookie.h
```diff
@@ -11,26 +11,43 @@
...
+
 #include <zeek-spicy/zeek-compat.h>
 
 namespace spicy::zeek::rt {
 
 namespace cookie {
 
+/** State stored inside protocol/file analyzer cookies to retain file analysis state. */
+struct FileState {
+    FileState(std::string analyzer_id) : analyzer_id(analyzer_id) {}
```
bbannier, line 24:
```suggestion
    FileState(std::string analyzer_id) : analyzer_id(std::move(analyzer_id)) {}
```

(Requires `#include <utility>`).

plugin/spicy/zeek_file.spicy
```diff
@@ -0,0 +1,28 @@
...
+    on %init {
+       zeek::file_begin(mime_type);
+
+       if ( size )
+           zeek::file_set_size(*size);
+    }
+
+    : bytes &chunked &eod { zeek::file_data_in($$); }
+
+    on %done { zeek::file_end(); }
```
bbannier, line 27: I believe we also need to terminate the feed on `%error`, i.e.,
```suggestion
    on %done { zeek::file_end(); }
    on %error { zeek::file_end(); }
```

plugin/spicy/zeek_file.spicy
```diff
@@ -0,0 +1,28 @@
...
+## size: Total number of bytes the file contains, if known; will be passed on to Zeek
+public type File = unit(mime_type: optional<string> = Null, size: optional<uint64> = Null) {
+    on %init {
+       zeek::file_begin(mime_type);
+
+       if ( size )
+           zeek::file_set_size(*size);
+    }
+
+    : bytes &chunked &eod { zeek::file_data_in($$); }
```
bbannier, line 25: I wonder whether we should move the part calling `file_data_in` into a filter consuming `bytes` so we do not have to store the full payload before being able to hand it over to Zeek. The machinery we are using should already support this.

bbannier (approved):

plugin/spicy/zeek_file.spicy
rsmmr: Good point, `%finally` will cover both.

plugin/spicy/zeek_file.spicy
rsmmr: `&chunked` takes care of that already actually.

bbannier (dismissed):

PR #36, opened 2021-05-06
Make enums' `Undef` available to Zeek scripts.

### LINKED ISSUE #35: referencing _Undef value for spicy enum in zeek code results in Aborted (core dumped)
The Spicy documentation section on [Zeek Integration - Enum Types](https://docs.zeek.org/projects/spicy/en/latest/zeek.html#enum-types) describes how Spicy enums are translated into code accessible on the Zeek side.

However, referencing the `_Undef` value on the zeek side results in a core dump:

`Test.spicy`
```
module Test;

public type MyEnum = enum {
    A = 83,
    B = 84,
    C = 85
};
```

`Test.zeek`
```
module Test;

event Test::message(c: connection) {
  local whatever: Test::MyEnum = Test::MyEnum_A;
  whatever = Test::MyEnum_Undef;
}
```

Output:
```
# zeek -NN /tmp/Test.hlto /host/Test.zeek 
internal error in /host/Test.zeek, line 5: enum value not found for Test::MyEnum_Undef
Aborted (core dumped)
```

Output commenting out the _Undef line results in no error.

rsmmr:
We can't represent the internal -1 on the Zeek side, so now using
INT_MAX instead.

Closes #35.


plugin/include/runtime-support.h
```diff
@@ -557,15 +557,19 @@ inline ::zeek::ValPtr to_val(const T& t, ::zeek::TypePtr target, const std::stri
...
  */
 template<typename T, typename std::enable_if_t<std::is_enum<T>::value>*>
 inline ::zeek::ValPtr to_val(const T& t, ::zeek::TypePtr target, const std::string& location) {
     if ( target->Tag() != ::zeek::TYPE_ENUM )
         throw TypeMismatch("enum", target, location);
 
-    return zeek::compat::EnumTypeGetEnumVal(target->AsEnumType(), static_cast<int>(t));
+    // Zeek's enum can't be negative, so we swap in max_int for our Undef (-1).
+    auto it = static_cast<int64_t>(t);
+    bro_int_t bt = (it >= 0 ? it : std::numeric_limits<::bro_int_t>::max());
```
bbannier, line 570: Hmm, it could be possible that a user specified an enum value with max_int for which we would now emit at least semantically incorrect code. It would be great if we could add some validation for that to declare that particular value "magic" and off-limits to users (ideally here, but I am not sure that kind of reflection is possible; else in Spicy proper).

rsmmr: I agree in principle, though in practice bro_int_t is a 64-bit integer and I'd be surprised if anybody ever defined that max as an enum value. Question is if we want to put this knowledge about bro_int_t into Spicy proper for a validation there.

bbannier: As a middle ground we could add a runtime check so that we do not silently mutilate data, e.g.,

```.cpp 
assert(((void)"Serializing enum values with value max_int to Zeek is not supported",
          it == std::numeric_limits<int64_t>::max()));
```

rsmmr: assert is debug mode only, I'll put a runtime exception in

bbannier: I lost track of what changed here since my initial comment, but the current code first does a possibly lossy conversion of `t: uint64_t` to `it: int64_t`, and in the creation of `bt: bro_int_t` checks for `it >= 0` which should then always be true.

Should we instead first check for max `int64_t` (and possibly throw), and then create `bt` (possibly performing the replacement `-1` -> `std::numeric_limits<::bro_int_t>::max()`)?

PR #37, opened 2021-05-25
Build package without parallelism.

bbannier:
Since building this package is pretty memory intensive, do not assume
that we can build in parallel. Users can override this by setting
`MAKEFLAGS` in the environment, e.g., to build using all CPUs:

    $ MAKEFLAGS="-j$(nproc)" zkg install zeek/spicy-analyzers

### DIFF
--- a/zkg.meta
+++ b/zkg.meta
@@ -2,7 +2,7 @@
 script_dir = plugin/scripts
 plugin_dir = build/plugin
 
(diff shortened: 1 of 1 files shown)


rsmmr (approved):
So not super-happy with this because it will slow down user installs significantly -- but I don' have a better idea either, so ok I guess.

PR #38, opened 2021-05-25
Update for current Zeek master

rsmmr:


plugin/include/zeek-compat.h
```diff
@@ -56,7 +56,9 @@
 #include <zeek/analyzer/Analyzer.h>
 #include <zeek/analyzer/Manager.h>
 #include <zeek/analyzer/protocol/tcp/TCP.h>
+#if ZEEK_VERSION_NUMBER < 40100 // Zeek < 4.1
```
bbannier, line 59: Aren't we including this same file on l.82 below (we actually already seem to do this for `TCP.h`).

Unless this file goes away in newer version let's just include it unconditionally.

bbannier (approved):

plugin/include/zeek-compat.h
rsmmr: it actually goes away

bbannier: Ah, missed how this added extra nesting to an existing `#if`.

PR #39, opened 2021-05-25
Add option Spicy::max_file_depth to limit maximum recursion of file a…

rsmmr:
…nalysis.

This avoids resource exhaustion with deeply nested files (potentionally
maliciously). Default is 5.

Note that this tracks Spicy-side nesting only, it doesn't count any
non-Spicy file analyzers (but we don't have any nested file analysis in
Zeek by default anyways).


tests/Baseline/zeek.file-analyzer-nested/weird.log
```diff
@@ -0,0 +1,11 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+#separator \x09
+#set_separator	,
+#empty_field	(empty)
+#unset_field	-
+#path	weird
+#open XXXX-XX-XX-XX-XX-XX
+#fields	ts	uid	id.orig_h	id.orig_p	id.resp_h	id.resp_p	name	addl	notice	peer	source
+#types	time	string	addr	port	addr	port	string	string	bool	string	string
+1363709111.086948	-	-	-	-	-	file_error	XXXXXXXXXXXXXXXXX: maximal file depth exceeded	F	zeek	-
```
bbannier, line 10: Since we do seem to report nothing in this case, do we actually have a chance to report any useful information from upper layers in the file? At least from the parsing side we have some interesting data, even before we hit the recursion limit.

bbannier (approved):

tests/Baseline/zeek.file-analyzer-nested/weird.log
rsmmr: Haven't tested it but II think we are actually continuing processing at the upper layer, just the new analyzer aborts directly. I'll see if I can confirm that.

rsmmr: Confired, and test extended accordingly.

PR #40, opened 2021-05-25
Extend `zeek_file::file_begin() to return the Zeek-side ID of the new file

rsmmr:
… file.


plugin/src/runtime-support.cc
```diff
@@ -259,12 +259,19 @@ std::string rt::fuid() {
...
 
-void rt::file_begin(const std::optional<std::string>& mime_type) {
+std::string rt::file_begin(const std::optional<std::string>& mime_type) {
     auto cookie = static_cast<Cookie*>(hilti::rt::context::cookie());
     auto* fstate = _file_state(cookie);
     ++fstate->file_id;
     fstate->mime_type = mime_type;
 
+    // Feed an empty chunk into the analysis to force creating the file state inside Zeek.
+    _data_in("", 0);
```
bbannier, line 269: With this we should now be able to remove the block at l.279.

bbannier (approved):

plugin/src/runtime-support.cc
rsmmr: oh, yeah, that's a rebase artifact, didn't get cleaned up.

bbannier (approved):

PR #43, opened 2021-06-08
Trigger a notice of file analyzer recursion limit.

### LINKED ISSUE #42: Zip analyzer event/notice on recursion limit
Feature Request.   It would be interesting to get a notice when the zip analyzer hits it's recursion limit.

bbannier:
Closes #42.


tests/zeek/file-analyzer-nested.zeek
```diff
@@ -24,6 +25,11 @@ event text::data3(f: fa_file, data: string)
 	print "data3", f$id, data;
 	}
 
+event Spicy::max_file_depth_exceeded(f: fa_file, args: Files::AnalyzerArgs, limit: count)
+	{
+	print "max file depth exceeded", f$id, args, limit;
```
rsmmr, line 30: This should be showing up `output-max`, but the `grep` is filtering it out, so that we don't have it the baseline; probably want to use some other text here not matching the grep.

Also, would be good check `notice.log` for the new `NOTICE`.

plugin/CMakeLists.txt
```diff
@@ -33,9 +33,8 @@ set(HEADERS
     zeek-reporter.h)
 
 set(ZEEK_SCRIPTS
-    scripts/__load__.zeek
-    scripts/base/spicy/__load__.zeek
-    scripts/base/spicy/main.zeek
+    scripts/_Zeek/spicy/__load__.zeek
```
rsmmr, line 36: This seems to work, though it isn't quite what I was hoping for. This is current structure:

```
# find scripts
scripts
scripts/misc
scripts/misc/record-spicy-batch.zeek
scripts/__load__.zeek
scripts/__preload__.zeek
scripts/_Zeek
scripts/_Zeek/spicy
scripts/_Zeek/spicy/main.zeek
scripts/_Zeek/spicy/__load__.zeek
scripts/_Zeek/__load__.zeek
```

I think this still has the problem of `scripts/__load__.zeek` being pulled in twice (because `script_dir` in `zkg.meta` hasn't changed); we just don't see that because it happens to be empty. We could just remove it, but IO think`script_dir` should really be pointing one level down. I was original hoping for this: set `script_dir` to `plugin/scripts/spicy`; then this layout:

```
scripts/spicy/__load__.zeek (contains @load ./base)
scripts/spicy/__preload__.zeek
scripts/spicy/base/main.zeek
scripts/spicy/base/__load__.zeek
scripts/spicy/misc/record-spicy-batch.zeek
```

 You might have tried this and it didn't work?

rsmmr:
Will need #46

PR #45, opened 2021-06-15
Experimental branch for building inside Zeek as a static plugin.

rsmmr:
Building & linking works for me, but it has script trouble at startup:

    > ./src/zeek
    fatal error: can't find base/init-bare.zeek


rsmmr:
Superseded by #46

PR #51, opened 2021-07-06
Make plugin version available to other projects.

rsmmr:


CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ set(SPICY_ZEEK_PLUGIN_VERSION_MAJOR ${CMAKE_MATCH_2})
 set(SPICY_ZEEK_PLUGIN_VERSION_MINOR ${CMAKE_MATCH_3})
 set(SPICY_ZEEK_PLUGIN_VERSION_PATCH ${CMAKE_MATCH_4})
 math(EXPR SPICY_ZEEK_PLUGIN_VERSION_NUMBER "${SPICY_ZEEK_PLUGIN_VERSION_MAJOR} * 10000 + ${SPICY_ZEEK_PLUGIN_VERSION_MINOR} * 100 + ${SPICY_ZEEK_PLUGIN_VERSION_PATCH}")
+set(SPICY_ZEEK_PLUGIN_VERSION_NUMBER "${SPICY_ZEEK_PLUGIN_VERSION_NUMBER}" CACHE STRING "")
```
bbannier, line 13: Like I already mentioned offline, that we had to `set` and already existing variable was surprising to me. Could you add a comment capturing our reasoning, e.g.,
```suggestion

# Additionally add the variable to the CMake cache so it is visible to other projects.
set(SPICY_ZEEK_PLUGIN_VERSION_NUMBER "${SPICY_ZEEK_PLUGIN_VERSION_NUMBER}" CACHE STRING "")
```

bbannier (dismissed):

PR #61, opened 2021-09-10
Extend `spicy_add_analyzer` to allow installing additional Zeek scripts.

### LINKED ISSUE #60: CMake function `spicy_add_analyzer` should support installing additional scripts
We provide `spicy_add_analyzer` to compile and install analyzers. We should add support to that script so accompanying Zeek scripts can be installed.

 See e.g., zeek/spicy-analyzers#77 for a related issue.

bbannier:
This patch changes `spicy_add_analyzer` to also support specifying
scripts which should be installed with the plugin. For that we change
the function to now support named arguments (`NAME`, `SOURCES`, and
`SCRIPTS`). For now we continue to support the previous format where
arguments were all unnamed, however this form does not support
installing scripts.

Closes #60.

### DIFF
--- a/cmake/ZeekSpicyAnalyzerSupport.cmake
+++ b/cmake/ZeekSpicyAnalyzerSupport.cmake
@@ -6,32 +6,69 @@
 
 include(GNUInstallDirs)
 
-# Add target to build an analyzer. Arguments are the name of the analyzer and a
-# variable number of source files for `spicyz`.
-function(spicy_add_analyzer name)
-    set(sources "${ARGN}")
-    string(TOLOWER "${name}" name_lower)
-    set(output "${SPICY_MODULE_OUTPUT_DIR_BUILD}/${name_lower}.hlto")
-    set(output_install "${SPICY_MODULE_OUTPUT_DIR_INSTALL}/${name_lower}.hlto")
-    set(deps "spicyz")
+# Add target to build an analyzer.
+#
+# Usage:
+#
+#     spicy_add_analyzer(
+#         NAME <analyzer_name>
+#         [SOURCES <source files for spicyz>...]
+#         [PACKAGE_NAME <package_name>]
+#         [SCRIPTS <additional script files to install>...]
+#     )
+function(spicy_add_analyzer)
+    set(options)
+    set(oneValueArgs NAME PACKAGE_NAME)
+    set(multiValueArgs SOURCES SCRIPTS)
+
+    cmake_parse_arguments(PARSE_ARGV 0 SPICY_ANALYZER "${options}" "${oneValueArgs}" "${multiValueArgs}")
+
+    # We also support the legacy behavior where the first arg is
+    # the analyzer NAME and all remaining arguments are SOURCES.
+    if ( SPICY_ANALYZER_UNPARSED_ARGUMENTS )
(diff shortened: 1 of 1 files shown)


cmake/ZeekSpicyAnalyzerSupport.cmake
```diff
@@ -6,32 +6,65 @@
...
+function(spicy_add_analyzer)
+    set(options)
+    set(oneValueArgs NAME)
+    set(multiValueArgs SOURCES SCRIPTS)
+
+    cmake_parse_arguments(PARSE_ARGV 0 SPICY_ANALYZER "${options}" "${oneValueArgs}" "${multiValueArgs}")
+
+    # We also support the legacy behavior where the first arg is
+    # the analyzer NAME and all remaining arguments are SOURCES.
+    if (SPICY_ANALYZER_UNPARSED_ARGUMENTS)
```
rsmmr, line 27: Formatting nit:

```suggestion
    if ( SPICY_ANALYZER_UNPARSED_ARGUMENTS )
```

Similar below.

cmake/ZeekSpicyAnalyzerSupport.cmake
```diff
@@ -6,32 +6,65 @@
...
+    add_custom_target(${SPICY_ANALYZER_NAME} ALL DEPENDS ${OUTPUT})
 
     if ( SPICY_MODULE_OUTPUT_DIR_INSTALL )
-        install(FILES ${output} DESTINATION "${SPICY_MODULE_OUTPUT_DIR_INSTALL}")
+        install(FILES ${OUTPUT} DESTINATION "${SPICY_MODULE_OUTPUT_DIR_INSTALL}")
+
+        if (DEFINED SPICY_ANALYZER_SCRIPTS)
+            install(
+                FILES ${SPICY_ANALYZER_SCRIPTS}
+                DESTINATION "${SPICY_SCRIPTS_OUTPUT_DIR_INSTALL}/spicy-analyzers/${NAME}")
```
rsmmr, line 62: This won't quite work yet: In the future we want to use this also for packages other than `spicy-analyzers`, so we shouldn't hardcode the name. Not immediately sure though what the solution is. Do we need another argument `PACKAGE_NAME` for this?

bbannier: Good catch. I went with an explicit `PACKAGE_NAME` argument which becomes required if `SCRIPTS` is also given.

cmake/ZeekSpicyAnalyzerSupport.cmake
```diff
@@ -6,32 +6,69 @@
...
-    string(TOLOWER "${name}" name_lower)
-    set(output "${SPICY_MODULE_OUTPUT_DIR_BUILD}/${name_lower}.hlto")
-    set(output_install "${SPICY_MODULE_OUTPUT_DIR_INSTALL}/${name_lower}.hlto")
-    set(deps "spicyz")
+# Add target to build an analyzer.
+#
+# Usage:
+#
+#     spicy_add_analyzer(
+#         NAME <name>
```
rsmmr, line 14:
```suggestion
#         NAME <analyzer_name>
```

rsmmr (approved):

PR #66, opened 2021-09-19
Add pre-commit hooks for CMake files.

bbannier:


rsmmr (commented):
Let's talk about this, not convinced yet that this is an improvement. At least we'd need to work on the configs, it's even changing indenting right now.

rsmmr (commented):
Ok, I generallive live with this :-) but would prefer the following changes (taking a guess at the options, haven't tried it):

- Leave comments alone (`enable_markup`)
- Insert space after if/else/endif (`separate_ctrl_name_with_space`)
- don't change variable spelling; looks like that's the linter, not automated; either leave that change out or we could instead prefix the lower-case ones with `_`, that way they'd be treated as internal by the formatter and ok to lower-case (guessing based on `local_var_pattern`)

bbannier:
Thanks for the great review and the suggestions @rsmmr. I applied your suggestions on comments and suggestions, and disabled the variable name convention check check.

cmake/ZeekSpicyAnalyzerSupport.cmake
```diff
@@ -49,40 +52,44 @@ function(spicy_add_analyzer)
...
     set_property(GLOBAL PROPERTY __spicy_included_analyzers "${tmp}")
-endfunction()
+endfunction ()
 
 # Flag that analyzer is *not* being built. This is purely informational:
 # the cmake output will contain a corresponding note. Arguments are the
 # name of the analyzers and a descriptive string explaining why it's
 # being skipped.
-function(spicy_skip_analyzer name reason)
+function (spicy_skip_analyzer name reason)
```
rsmmr, line 85: This formatting is a a bit odd, but I guess there's nothing to do about it with `separate_ctrl_name_with_space` on.

rsmmr (commented):
Sorry, I got one more request. I find changes like these to be pretty verbose:

```
- set(SPICY_CONFIG "${spicy_config}" CACHE FILEPATH "")
+ set(SPICY_CONFIG
             "${spicy_config}"
             CACHE FILEPATH "")
```

I looks like setting  `max_subgroups_hwrap` to 3 instead of the default 2 might help, mind trying?

cmake/ZeekSpicyAnalyzerSupport.cmake
bbannier: Yes, the formatter doesn't really seem to distinguish these since in CMake on the AST level `if ... endif` probably isn't much different from `function ... endfunction`.

bbannier:
> Sorry, I got one more request. I find changes like these to be pretty verbose:
> 
> ```
> - set(SPICY_CONFIG "${spicy_config}" CACHE FILEPATH "")
> + set(SPICY_CONFIG
>              "${spicy_config}"
>              CACHE FILEPATH "")
> ```
> 
> I looks like setting `max_subgroups_hwrap` to 3 instead of the default 2 might help, mind trying?

Yes, that seems to subjectively lead to fewer of these cases. Added that.

rsmmr (approved):

PR #68, opened 2021-09-20
Add clang-tidy CI job.

### LINKED ISSUE #18: Add clang-tidy to CI
Currently we check the code through Spicy's CI, but would be good to catch trouble earlier.

bbannier:
Closes #18.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -83,6 +83,9 @@ standalone_ubuntu_task:
   build_script:
     - (mkdir build && cd build && cmake -DCMAKE_INSTALL_PREFIX=/opt/spicy-plugin -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -G Ninja .. && ninja -j4)
 
+  clang_tidy_script:
+    - (cd build && run-clang-tidy -j $(nproc) $(jq -r '.[].file | select(contains(".bif") | not)' <compile_commands.json ))
+
   test_build_script:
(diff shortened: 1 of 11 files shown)


.cirrus.yml
```diff
@@ -81,7 +81,7 @@ standalone_ubuntu_task:
     - rm spicy-dev.deb
 
   build_script:
-    - (mkdir build && cd build && cmake -DCMAKE_INSTALL_PREFIX=/opt/spicy-plugin -G Ninja .. && ninja -j4)
+    - (mkdir build && cd build && cmake -DCMAKE_INSTALL_PREFIX=/opt/spicy-plugin -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -G Ninja .. && ninja -j4)
```
rsmmr, line 84: The commit says "if possible": does this mean that cmake will ignore the `CMAKE_CXX_COMPILER_LAUNCHER` setting if `ccache` isn't found? Or will it bail out? (Just wondering, either way is fine I think because we control the setting and know its installed)

rsmmr (approved):

.cirrus.yml
bbannier: We currently do not set up ccache e.g., for installations via `zkg`. This should be also be possible (e.g., by replace the compiler invoked by e.g., `spicyc` with a symlink to `ccache` in the container image), but I didn't attempt it here.

rsmmr: Ok, probably just misunderstood your comment in the commit message.

PR #83, opened 2021-12-09
Provide API back into Zeek's protocol analysis.

### LINKED ISSUE #76: Provide primitives so that Spicy analyzers can forward data to other Zeek analyzers for follow-up analysis
We provide functions to pass data extracted by Spicy file analyzers back into Zeek with e.g.,
```.zeek
function zeek::file_begin(mime_type: optional<string> = Null) : string
```

Such functionality currently does not exist for Spicy protocol or packet analyzers. We should provide similar functions for such analyzers, ideally with some optional type so callers can opt into a specific analyzer to process the data (e.g., for HTTP with websockets can detect that the unparsed data is websockets) if they do not want Zeek to detect the format dynamically.

rsmmr:
Closes #76.


src/runtime-support.cc
```diff
@@ -231,6 +235,74 @@ void rt::reject_protocol(const std::string& reason) {
...
+            // AddChildAnalyzer() will have deleted child
+            throw ZeekError(::hilti::rt::fmt("could not add analyzer '%s' to connection", *analyzer));
+    }
+    else {
+        // Use a Zeek PIA stream analyzer performing DPD.
+        auto child = new ::zeek::analyzer::pia::PIA_TCP(c->analyzer->Conn());
+
+        if ( ! c->analyzer->AddChildAnalyzer(child) )
+            // AddChildAnalyzer() will have deleted child
+            throw ZeekError(::hilti::rt::fmt("could not add DPD analyzer to connection"));
```
bbannier, line 261: This leaks `child`. Let's either clean it up here, it instead use some smart pointer to hold it.

rsmmr: See the comment. :-)

bbannier: 🤦

bbannier (approved):

PR #84, opened 2021-12-09
Wrap different Zeek tag types

### LINKED ISSUE #81: Uses of `Tag`s incompatible with Zeek `master` branch
[This commit](https://github.com/zeek/zeek/commit/331161138a28b633239dbc2eb98ab6ccafe47076) on Zeek's `master` branch reorganized the different `Tag` types and their header files. We cannot build against this commit or any newer version.

The following issues need to be fixed:

- [ ] fix includes so we always pull in headers declaring a specific used (now deprecated) `Tag` variant.
- [ ] rewrite our `toggleAnalyzer` functions. Since all different `Tag` types now are just aliases to a single `::zeek::Tag` we cannot use overloads here anymore. We also need to adjust the logic in the "frontend" `toggleAnalyzer` function so it branches to the correct concrete functions (alternatively: could we just get rid of this function?).

bbannier:
As of zeek-4.2 `Tag` types in Zeek got unified. This leads to issues for
us since we e.g., overload by `Tag` type, or need to be able to handle
individual `Tag` types differently.

This patch introduces wrapper types around Zeek `Tag` types.

Closes #81.

### DIFF
--- a/include/zeek-compat.h
+++ b/include/zeek-compat.h
@@ -40,6 +40,16 @@
 #include <zeek/packet_analysis/Analyzer.h>
 #endif
 
+#if ZEEK_VERSION_NUMBER < 40200 // Zeek < 4.2
+#include <zeek/analyzer/Tag.h>
+#include <zeek/file_analysis/Tag.h>
+#if ZEEK_VERSION_NUMBER >= 30300 // Zeek >= 3.3 (aka 4.0)
+#include <zeek/packet_analysis/Tag.h>
+#endif
+#else
+#include <zeek/Tag.h>
+#endif
+
 #if ZEEK_VERSION_NUMBER >= 30200 // Zeek >= 3.2
 #include <zeek/Conn.h>
 #include <zeek/DebugLogger.h>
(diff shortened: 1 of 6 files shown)


include/zeek-compat.h
```diff
@@ -200,6 +210,42 @@ using ::attr_tag::ATTR_OPTIONAL;
 
 //// Wrapper functions for functionality that differs by version.
 
+namespace spicy::zeek::compat {
+#if ZEEK_VERSION_NUMBER < 40200 // Zeek < 4.2
+using AnalyzerTag = ::zeek::analyzer::Tag;
+using FileAnalysisTag = ::zeek::file_analysis::Tag;
```
rsmmr, line 216: For types, the existing code in `zeek-compat.h` tries to make them available to old Zeek versions by declaring them under their new Zeek name (in contrast to putting them into the `compat` namespace). Can we do that here too?

bbannier: I agree, this would be the right thing to do for long-term maintenance.

It would however be a bigger change, e.g., from a quick look I saw:

- need to provide wrappers for many functions taking or returning `Tag` types so we can convert old `Tag`s into the new `Tag`
- need to provide wrappers for functions of `Tag`, e.g., `AsVal` previously had a different signature

This list could change while I go through the code. I'll try that.

rsmmr: Ok, not worth if it's signficant effort.

bbannier: I took a stab at this, and just using `::zeek::Tag` in all versions doesn't really work since it's API changed (e.g., before 4.2 it has no default constructor), so we'd have to come up with our own type anyway. I'd suggest we stay what is in this PR.

rsmmr (approved):

Issue #91, opened 2022-01-21
Allow multiple calls to `protocol_begin()`


rsmmr:
Currently we abort if `protocol_begin()` gets called twice without any `protocol_end()` in between. While that's reasonable in principle, it can make it hard to call `protocol_begin()` at the right time when both originator and responder side may want to trigger it. In that case, currently one gets an obscure error message of Zeek not being able to add DPD analyzer.

I believe it should be fine to just ignore the 2nd call silently in these cases. 

(Note, I believe `protocol_end()` is already safe against calling multiple times, but should double check).

PR #93, opened 2022-01-28
Support forwarding to Zeek TCP-level application data from a UDP anal…

### LINKED ISSUE #92: Use of `protocol_*` functions in corelight/zeek-spicy-openvpn triggers sigabort
As of corelight/zeek-spicy-openvpn@6bcb4ec6dd3c8814cbb5186e688585b7bf105f97 the test suite of that analyzer fails on their Linux CI. The logs mention that many tests failed due to `SIGABRT`. This seems to coincide with them starting to use `protocol_*` functions from this package. 

We should look into this and depending on what we find either fix our implementation, or reject incorrect usage patterns.

This is also tracked in corelight/zeek-spicy-openvpn#3.

rsmmr:
…yzer.

This necessarily remains a bit hackish because of a lack of better
Zeek-side APIs: we have to create a fake TCP analyzer that's just
there for our actual child analyzer to have one to access; no further
TCP semantics are provided.

This comes with a new check that any connected analyzer is indeed a
TCP application-layer analyzer. We were already limited to that case,
but didn't catch it so far. We could open things up to UDP in the
future as well, but would need a way for the user to select which one
to use. I'll leave that for later once a use-case comes up.

Closes #92.


rsmmr:
reverting to draft, I'll add #91 on top here.

bbannier (approved):
LGTM. This seems to fix corelight/zeek-spicy-openvpn#3 which seems to only require a baseline update.

PR #96, opened 2022-03-01
Add support for working on partial data or content with gaps

bbannier:
This adds the plugin-side work for zeek/spicy#1136.


tests/Baseline/zeek.gap-recovery/output
```diff
@@ -0,0 +1,3 @@
+### BTest baseline data generated by btest-diff. Do not edit. Use "btest -U/-u" to update. Requires BTest >= 0.63.
+[$responses=[[$marker=b"HTTP", $len=3070], [$marker=b"HTTP", $len=3070], [$marker=b"HTTP", $len=21999], [$marker=b"HTTP", $len=3070], [$marker=b"HTTP", $len=21999], [$marker=b"HTTP", $len=3070], [$marker=b"HTTP", $len=21999]]]
```
rsmmr, line 2: Do I see it right that the gap is inside the 2nd response? (just checking if I'm reading things right)

rsmmr (approved):

tests/Baseline/zeek.gap-recovery/output
bbannier: Yes, the second response is missing.

PR #100, opened 2022-03-23
Prepare plugin for optional direct inclusion into Zeek.

bbannier:


rsmmr (approved):
I assume this means that we will inject everything that `configure`  would normally set from Zeek's cmake?

bbannier:
> I assume this means that we will inject everything that `configure` would normally set from Zeek's cmake?

Yes, exactly. Currently this seems sufficient: https://github.com/zeek/zeek/blob/8d28cfa320143d603ac8facc42c3ac8d2b99c638/CMakeLists.txt#L371-L384.

bbannier:
Thanks for the review @rsmmr. I'll put this back into draft mode and will keep pushing changes needed for Zeek integration (zeek/zeek#2006) to this branch.

cmake/FindSpicy.cmake
```diff
@@ -11,17 +11,10 @@
 #
 #     If SPICY_FOUND is true:
 #
-#         SPICYC                        full path to spicyc
```
rsmmr, line 14: I'm just realizing something: analyzers are pulling this in indirectly through `find_package(Spicy)` in `FindSpicyPlugin.cmake`. Could they be using any of the removed values?

cmake/FindSpicy.cmake
```diff
@@ -141,12 +130,8 @@ function (spicy_print_summary)
             "\nFound Spicy:           ${HAVE_SPICY}")
 
     if (HAVE_SPICY)
-        message(
-            "\nVersion:               ${SPICY_VERSION} (${SPICY_VERSION_NUMBER})"
-            "\nPrefix:                ${SPICY_PREFIX}"
```
rsmmr, line 146: This one would actually be useful to keep for debugging what Spicy installation is being used. We could just set it to some descriptive string when building from inside Zeek.

cmake/FindSpicy.cmake
bbannier: Yes, they might be. Should I drop the cleanup patch?

rsmmr: Well, yes, but when building from Zeek, we then actually need to set them to appropriate values too.

cmake/FindSpicy.cmake
bbannier: I will keep this part as is.

rsmmr (approved):

PR #105, opened 2022-04-08
Added close_connection capability

luxanna91:
Added close_connection function accessible from spicy.
E.g.: to close long UDP connections.

### DIFF
--- a/include/runtime-support.h
+++ b/include/runtime-support.h
@@ -271,7 +271,7 @@ std::string fuid();
 /**
  * Close the current active connection.
  */
-void close_connection();
+void terminate_session();
 
 /**
  * Signals the expected size of a file to Zeek's file analysis.


include/runtime-support.h
```diff
@@ -268,6 +268,11 @@ std::string file_begin(const std::optional<std::string>& mime_type);
  */
 std::string fuid();
 
+/**
+ * Close the current active connection.
+ */
+void close_connection();
```
rsmmr, line 274: How about calling this `terminate_session` to go with the more general terminology that the session manager brings these days?

src/runtime-support.cc
```diff
@@ -444,6 +445,22 @@ static void _data_in(const char* data, uint64_t len, std::optional<uint64_t> off
     }
 }
 
+void rt::close_connection(){
+    auto cookie = static_cast<Cookie*>(hilti::rt::context::cookie());
+    assert(cookie);
+
+    if ( auto c = std::get_if<cookie::ProtocolAnalyzer>(cookie) ) {
+        if (::zeek::session_mgr){
```
rsmmr, line 453: The session manager is always available in Zeek, just so an `assert(::zeek::session_mgr)` should do it here.

tests/zeek/close-connection.spicy
```diff
@@ -0,0 +1,30 @@
+# @TEST-EXEC: spicyz -o test.hlto %INPUT test.evt
+# @TEST-EXEC: ${ZEEK} -b -r ${TRACES}/long-dns-connection.pcap _Zeek::Spicy test.hlto
```
rsmmr, line 2: This is not actually checking that `close_connection()` indeed has the intended effect. I think this should generate multiple entries in `conn.log` now instead of just one as it would have previously done, right? So could do a `btest-diff` on `conn.log` to confirm that.

rsmmr (commented):
This looks good to me, just some small comments. I can also just tweak it a bit during merge if you prefer.

luxanna91:
Hi, thanks for your answer. You can tweak it during merge for sure. Calling it `terminate_session` sounds good to me.

rsmmr:
Superseded by #107.

rsmmr:
@luxanna91 just for the record: where's that trace coming from, did you create that yourself?

luxanna91:
@rsmmr I got it from wireshark pcap examples [page](https://wiki.wireshark.org/SampleCaptures), [direct download link ](https://wiki.wireshark.org/uploads/__moin_import__/attachments/SampleCaptures/dns.cap)

PR #108, opened 2022-04-13
Add support for passing Spicy structs to Zeek as records.

### LINKED ISSUE #95: Add support for passing Spicy structs to Zeek as records.
This conversion is currently missing. It would work similarly to tuples.

bbannier:
Closes #95.


include/runtime-support.h
```diff
@@ -629,6 +631,64 @@ inline ::zeek::ValPtr to_val(const T& t, ::zeek::TypePtr target, const std::stri
...
+    int idx = 0;
+
+    auto numFields = rtype->NumFields();
+
+    t.__visit([&](const auto& name, const auto& val) {
+        if ( idx >= numFields )
+            throw TypeMismatch(hilti::rt::fmt("no matching record field for field '%s'", name), location);
+
+        auto field = zeek::compat::RecordType_GetFieldType(rtype, idx);
+        std::string fieldName = rtype->FieldName(idx);
```
rsmmr, line 655: `field_name`

include/runtime-support.h
```diff
@@ -629,6 +631,64 @@ inline ::zeek::ValPtr to_val(const T& t, ::zeek::TypePtr target, const std::stri
...
+inline ::zeek::ValPtr to_val(const T& t, ::zeek::TypePtr target, const std::string& location) {
+    if ( target->Tag() != ::zeek::TYPE_RECORD )
+        throw TypeMismatch("struct", target, location);
+
+    auto rtype = target->AsRecordType();
+
+    auto rval = std::make_unique<::zeek::RecordVal>(zeek::compat::ToValCtorType(rtype));
+    int idx = 0;
+
+    auto numFields = rtype->NumFields();
```
rsmmr, line 648: `num_fields`

include/runtime-support.h
```diff
@@ -629,6 +631,64 @@ inline ::zeek::ValPtr to_val(const T& t, ::zeek::TypePtr target, const std::stri
...
+                               location);
+
+        ::zeek::ValPtr v = nullptr;
+
+        if constexpr ( std::is_same<decltype(val), const hilti::rt::Null&>::value ) {
+            // "Null" turns into an unset optional record field.
+        }
+        else
+            // This may return a nullptr in cases where the field is to be left unset.
+            v = to_val(val, field, location);
```
rsmmr, line 668: Does that include optionals? An unset Spicy-optional should turn into an unset Zeek-side record field. (I know this code is copied from above, but not sure how optionals are handled.)

include/runtime-support.h
```diff
@@ -629,6 +631,64 @@ inline ::zeek::ValPtr to_val(const T& t, ::zeek::TypePtr target, const std::stri
...
+        }
+
+        idx++;
+    });
+
+    // We already check above that all Spicy-side fields are mapped so we
+    // can only hit this if there are uninitialized Zeek-side fields left.
+    if ( idx != numFields )
+        throw TypeMismatch(hilti::rt::fmt("missing initialization for field '%s'", rtype->FieldName(idx + 1)),
+                           location);
```
rsmmr, line 687: I think this good enough as it is, but just to say it, we could be more fancy as well: we could (1) match fields by name so that the struct and record types don't need to have the same ordering, and (2) we could allow fields to be missing if the corresponding record field was optional/defaulted. Implementing that would require more bookkeeping, probably not worth it for now; can always generalize  later.

include/runtime-support.h
```diff
@@ -629,6 +631,64 @@ inline ::zeek::ValPtr to_val(const T& t, ::zeek::TypePtr target, const std::stri
...
+        throw TypeMismatch("struct", target, location);
+
+    auto rtype = target->AsRecordType();
+
+    auto rval = std::make_unique<::zeek::RecordVal>(zeek::compat::ToValCtorType(rtype));
+    int idx = 0;
+
+    auto numFields = rtype->NumFields();
+
+    t.__visit([&](const auto& name, const auto& val) {
```
rsmmr, line 650: Neat trick to use `__visit` here!

rsmmr (approved):

include/runtime-support.h
bbannier: Yes, this seems to largely work for optional fields. I'll update the test for that.

While testing this I also found a bug in how we handle unset `attrs` below where we would have segfaulted on an unset `&optional` Spicy-side field mapped onto a non-`&optional` Zeek-side field. I'll append a patch fixing this here and in the handling of tuples.

rsmmr (approved):

PR #114, opened 2022-05-31
Mark spicy_analyzer_for_mime_type with &is_used

timwoj:
This is related to https://github.com/zeek/zeek/pull/2090. Without this change, all of Zeek's btests complain about this method not being used.


scripts/Zeek/Spicy/bare.zeek
```diff
@@ -48,7 +48,9 @@ export {
 # doc-functions-end
 }
 
-event spicy_analyzer_for_mime_type(a: Files::Tag, mt: string)
+# Marked with &is_used to suppress complaints when there aren't any
+# Spicy file analyzers loaded, and hence this event can't be generated.
+event spicy_analyzer_for_mime_type(a: Files::Tag, mt: string) &is_used
```
bbannier, line 53: Since this attribute is a syntax error if it is not known, let's guard this with a version check on e.g., 5.0.0, see e.g., https://github.com/zeek/spicy-plugin/blob/248f839a0f419af0b56f781443f4adc244f08329/scripts/Zeek/Spicy/bare.zeek#L29.

PR #124, opened 2022-07-11
Do not install test scripts.

bbannier:


bbannier:
Not needed as these files are installed for reuse by other packages.

PR #127, opened 2022-07-12
Add support for adding additional libraries to link

### LINKED ISSUE #125: Add support for adding additional libraries to link
zeek/spicy#1244 adds a command line flag which can be used add additional libraries to the link. We should make that command line flag accessible in `spicy_add_analyzer`.

bbannier:
Closes #125.

### DIFF
--- a/src/compiler/bin/spicyz.cc
+++ b/src/compiler/bin/spicyz.cc
@@ -13,9 +13,12 @@ const ::hilti::logging::DebugStream ZeekPlugin("zeek");
 
 void ::spicy::zeek::debug::do_log(const std::string& msg) { HILTI_DEBUG(ZeekPlugin, std::string(msg)); }
 
+constexpr int OPT_CXX_LINK = 1000;
+
 static struct option long_driver_options[] = {{"abort-on-exceptions", required_argument, nullptr, 'A'},
                                               {"show-backtraces", required_argument, nullptr, 'B'},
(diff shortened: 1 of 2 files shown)


src/compiler/bin/spicyz.cc
```diff
@@ -196,6 +200,14 @@ static hilti::Result<Nothing> parseOptions(int argc, char** argv, hilti::driver:
 
             case 'z': std::cout << spicy::zeek::configuration::ZeekConfig << std::endl; return Nothing();
 
+            case OPT_CXX_LINK:
+#if SPICY_VERSION_NUMBER >= 10600
+                compiler_options->cxx_link.emplace_back(optarg);
+#else
+                return hilti::result::Error("option '--cxx-link' is only supported for >=spicy-1.6");
```
rsmmr, line 207: Nit: 
```suggestion
                return hilti::result::Error("option '--cxx-link' is only supported for >= Spicy 1.6");
```

rsmmr (dismissed):

src/compiler/bin/spicyz.cc
bbannier: If we are going for prose I'll use

```cpp
return hilti::result::Error("option '--cxx-link' is only supported for Spicy 1.6 or newer");
```

rsmmr: Works for me. :-)

PR #138, opened 2022-10-06
Fix analyzer replacement.

### LINKED ISSUE #137: zeek/spicy: Built-in spicy modules cannot be replaced.
When Zeek is built with built-in spicy modules and a custom spicy module is loaded that uses the `replaces` keyword for an existing spicy module, the existing spicy module is not replaced -- it's left loaded along with the custom spicy module which requested the replacement.

rsmmr:
Replacing a Zeek-side analyzer didn't work if (a) the Spicy plugin was built into Zeek (like Zeek does by default now), (b) the plugin was initialized after that Zeek-side analyzer (which happened with names lexicographically sorting after "Spicy"), and (c) the HLTO was loaded from the module search path (instead of from the command line).

For regression testing, this changes one of our existing replacement tests to load the HLTO through the search path.

Closes #137.


include/plugin.h
```diff
@@ -253,6 +253,9 @@ class Plugin : public zeek::plugin::Plugin {
     // Return a Zeek location object for the given file name that will stay valid.
     ::zeek::detail::Location makeLocation(const std::string& fname);
 
+    // Disable any Zeek-side analyzers that are replaced by on of ours.
```
bbannier, line 256:
```suggestion
    // Disable any Zeek-side analyzers that are replaced by one of ours.
```

tests/zeek/replaces.zeek
```diff
@@ -1,5 +1,9 @@
-# @TEST-EXEC: ${ZEEK} -r ${TRACES}/ssh-single-conn.trace  ssh.spicy ./ssh.evt %INPUT | sort >output
+# @TEST-EXEC: mkdir -p modules
+# @TEST-EXEC: spicyz -o modules/ssh.hlto ssh.spicy ./ssh.evt
+# @TEST-EXEC: ZEEK_SPICY_MODULE_PATH=$(pwd)/modules ${ZEEK} -r ${TRACES}/ssh-single-conn.trace %INPUT | sort >output
 # @TEST-EXEC: btest-diff output
+#
+# We use the module search path for loading here as a regression test for #137.
```
bbannier, line 6: I was able to confirm that the changes lead to correct behavior by testing in our Docker image (with patched spicy-plugin).

I tried this test without the plugin changes, and at least on my machine (macos, Zeek and Spicy from Homebrew) this test always succeeds.

bbannier (dismissed):

src/plugin.cc
```diff
@@ -725,3 +700,46 @@ void plugin::Zeek_Spicy::Plugin::autoDiscoverModules() {
...
+        if ( ! tag ) {
+            ZEEK_DEBUG(hilti::rt::fmt("%s is supposed to replace protocol analyzer %s, but that does not exist",
+                                      info.name_analyzer, info.name_replaces));
+
+            continue;
+        }
+
+        ZEEK_DEBUG(hilti::rt::fmt("  Replaces existing protocol analyzer %s", info.name_replaces));
+        info.replaces = tag;
+        ::zeek::analyzer_mgr->DisableAnalyzer(tag);
```
bbannier, line 719: Just as a note, before this change we would potentially replace less analyzers than expected while now we should always disable the replaced analyzer. That makes it possible to e.g., have two Spicy analyzers which replace each other so that with this impl none would be active. We could add a validation step to reject that.

Same below.

tests/zeek/replaces.zeek
rsmmr: I had tested this test locally without the change, and it failed as expected for me.

src/plugin.cc
rsmmr: Seems fine to me to leave as is.

tests/zeek/replaces.zeek
bbannier: Yepp, this only fails if one uses a builtin spicy-plugin. Maybe we should add a comment.

bbannier (approved):

PR #139, opened 2022-10-07
Allow to use Zeek-side `ANALYZER_*` tags in JIT mode.

rsmmr:
We had an inconsistency between pre-compiled HLTO files and JITing an analyzer by giving Zeek the `*.{evt,spicy}` files directly: in the latter case, a Zeek script wasn't able to use the `ANALYZER_*` tags, because they weren't defined at script load time yet. The reason is that they got registered only once the HILTI-side module initialization ran, which for JITed analyzers is later than for HLTOs.

We can't really change when module initialization executes, so this commit fixes the issue differently: when running from inside Zeek, we now register the analyzer with Zeek immediately once we read its definition from the EVT file; turns out we have all information at that point that we need. We then still run the module initialization later and just ignore the additional registration as long as they match.

Note that there's still an ordering requirement: the EVT file needs to be loaded before the Zeek script using the tag. However, that's the same for HLTOs, so that's consistent at least.

This is on top of #138.


rsmmr:
@bbannier is your sign-off for this included in #141?

bbannier:
> @bbannier is your sign-off for this included in #141?

Yes, that was the intention. Sorry for not being explicit.

bbannier (approved):

PR #147, opened 2022-10-24
Tie plugin to Spicy version, and analyzers to plugin.

rsmmr:
This uses versions-encoded-in-function names to (1) make sure the plugin gets rebuild if the Spicy version changes, and (2) analyzers get rebuild if the plugin versions changes. Through a separate Spicy-side changes, analyzers are in additional also tied to the Spicy version directly (https://github.com/zeek/spicy/pull/1282)

### DIFF
--- a/src/plugin.cc
+++ b/src/plugin.cc
@@ -75,6 +75,14 @@ private:
 };
 
 plugin::Zeek_Spicy::Plugin::Plugin() {
+#ifdef HILTI_VERSION_FUNCTION
+    // This ensures version compatibility at dlopen() time  by requiring the
+    // versioned symbol to be present. The symbol is available starting with
+    // Spicy 1.6.
+    _spicy_version = HILTI_VERSION_FUNCTION();
+#endif
+
+    // Not sure if with the with the check above, we still need this? Can't hurt I guess.
     if ( spicy::zeek::configuration::ZeekVersionNumber != ZEEK_VERSION_NUMBER )
         reporter::fatalError(
             hilti::rt::fmt("Zeek version mismatch: running with Zeek %d, but plugin compiled for Zeek %s",
--- a/src/plugin.cc
+++ b/src/plugin.cc
@@ -33,6 +33,8 @@ const hilti::logging::DebugStream ZeekPlugin("zeek");
 }
 #endif
 
+const char* ZEEK_SPICY_PLUGIN_VERSION_FUNCTION() { return spicy::zeek::configuration::PluginVersion; }
+
 plugin::Zeek_Spicy::Plugin SpicyPlugin;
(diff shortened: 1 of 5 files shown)


rsmmr:
Waiting for https://github.com/zeek/spicy/pull/1282 to land before proceeding.

src/plugin.cc
```diff
@@ -75,6 +77,11 @@ class OurGlueCompiler : public spicy::zeek::GlueCompiler {
 };
 
 plugin::Zeek_Spicy::Plugin::Plugin() {
+    // This ensures version compatibility at dlopen() time  by requiring the
+    // versioned symbol to be present.
+    _spicy_version = HILTI_VERSION_FUNCTION();
```
bbannier, line 82: This seems to be the only aspect requiring >=spicy-1.6.0 and I see you already pulled that into its own commit. Just wondering, do we want to backport the other two changes into the spicy-plugin version we pull into the Zeek LTS release?

src/plugin.cc
```diff
@@ -75,6 +77,11 @@ class OurGlueCompiler : public spicy::zeek::GlueCompiler {
...
+
+    // Not sure if with the with the check above, we still need this? Can't hurt I guess.
```
bbannier, line 84: This seems to check another compatibility aspect, so this should either already have been handled by the Zeek version functionality or we should instead evaluate that symbol like we do with `HILTI_VERSION_FUNCTION`.

bbannier (dismissed):

src/plugin.cc
bbannier: Actually wondering if we shouldn't make this conditional on the Spicy version we are running against (e.g., on the preprocessor level). This should simplify what can go into Zeek. As is we cannot publish any version of the plugin until spicy-1.6.0 is released.

rsmmr: I'd say let's not backport into the Zeek release branch. But I can put just an `#ifdef` around this, then we can decide what/when to pull upstream.

bbannier: Sounds good, that should also allow us running CI against more Spicy versions.

src/plugin.cc
rsmmr: Not sure what you mean by "evaluate that symbol like we do with HILTI_VERSION_FUNCTION"?

bbannier: I am unsure whether the symbol is evaluated automatically with the Zeek-side tooling we are using, or whether we need code like `_spicy_version = HILTI_VERSION_FUNCTION();` but for the Zeek version function.

rsmmr: Ah, Zeek's plugin integration takes care of calling the Zeek version function.

PR #148, opened 2022-10-30
cmake replace quote string

nabbi:
CMake Error
string sub-command REPLACE requires at least four arguments.

cmake version 3.24.2


bbannier:
Merged into both `main` and `release/1.3` and available in `v1.4.1` and `v1.3.22`.

PR #149, opened 2022-11-02
Create Zeek-side record types automatically

rsmmr:


src/compiler/driver.cc
```diff
@@ -343,11 +328,31 @@ hilti::Result<hilti::Nothing> Driver::compile() {
...
+
+    std::vector<TypeInfo> result;
+    result.reserve(_types.size());
+    for ( const auto& t : _types ) {
+        if ( exported_only && std::find_if(exported.begin(), exported.end(),
+                                           [&t](const auto& i) { return i.first == t.first; }) == exported.end() )
+            continue;
+
+        result.push_back(t.second);
+    }
```
bbannier, line 349: Worst case this will be a quadratic loop in the number of types.

For the `exported_only` case we could simply iterate over `exported` and lookup `TypeInfo`s in `_types`. We could add a branch doing that. In that case we should also adjust the `reserve` to only ask for `exported.size()`.

src/plugin.cc
```diff
@@ -306,32 +312,43 @@ void plugin::Zeek_Spicy::Plugin::registerPacketAnalyzer(const std::string& name,
...
+    auto type = createZeekType(ti->type, id);
+    if ( ! type )
+        return type.error();
+
+    registerType(ti->id, *type);
+    return hilti::Nothing();
+}
+
+void plugin::Zeek_Spicy::Plugin::registerType(const hilti::ID& id, const ::zeek::TypePtr& type) {
+    if ( ::zeek::detail::lookup_ID(id.str().c_str(), id.namespace_().str().c_str()) ) {
```
bbannier, line 329: Since we now provide users an explicit way to inject input into this method, I wonder whether we shouldn't instead throw an error if the same `id` is registered multiple times (maybe check whether `type` is inconsistent?). The defensive pattern would be to always first call `findType` and only if it finds nothing to register a new type.

src/plugin.cc
```diff
@@ -942,3 +959,199 @@ const ::zeek::plugin::Component* plugin::Zeek_Spicy::Plugin::findComponent(const
...
+
+        if ( auto t = zeek_types.find(id); t != zeek_types.end() ) {
+            if ( t->second )
+                return t->second;
+            else
+                return hilti::result::Error(hilti::util::fmt("type '%s' is self-recursive", id));
+        }
+        else
+            return hilti::result::Error(
+                hilti::util::fmt("type requires another type that is not (yet) exported (%s)", id));
```
bbannier, line 1009: Would be nice to add a test for this.

src/plugin.cc
bbannier: Just noticed the debug statement added in a later patch. Maybe that is good enough, but I suspect you added this because you ran into exactly the issue a hard error had prevented.

tests/zeek/export-types.zeek
```diff
@@ -0,0 +1,105 @@
...
+    for ( i in all_globals ) {
+	id = all_globals[i];
+
+	if ( ! (/Test::/ in id) )
+	    next;
+
+	if ( /type_record_/ in id )
+            print id, record_fields(id);
+	# else if ( /type_enum$/ in id )
+	#     print id, enum_names(id); # Not available in 5.0 yet
```
bbannier, line 55: Intentionally left commented out?

bbannier (commented):
In addition to the syntax here I was also hoping for support for extended syntax like `export X::Y as A::B` (or even export into the global Zeek namespace); this would allow tailoring where some (possibly not even user-controlled) type ends up. It seems the infrastructure added in this PR could be largely adjusted for that, but this might also introduce new scenarios for name collisions or other errors.

rsmmr:
Moving back to draft, this still has some issues.

rsmmr:
This is waiting for #154 now.

tests/zeek/export-types.zeek
rsmmr: Yes, to be activated once we don't need to support 5.0 anymore.

src/plugin.cc
rsmmr: This is gone now without the JIT code.

src/compiler/driver.cc
rsmmr: I'll leave this as is, `exported_types` is going to remain tiny.

rsmmr:
>  `export X::Y as A::B` 

Ack, I'm adding this.

src/compiler/driver.cc
rsmmr: Well, turns out I'm moving the `exported_only` cases into a separate functions, which then indeed ends up iterating over `exported` to ensure a consistent order.

rsmmr:
This is ready (again) now.

src/compiler/driver.cc
```diff
@@ -343,11 +328,43 @@ hilti::Result<hilti::Nothing> Driver::compile() {
...
+    return result;
+}
+
+std::vector<std::pair<TypeInfo, hilti::ID>> Driver::exportedTypes() const {
+    std::vector<std::pair<TypeInfo, hilti::ID>> result;
+
+    for ( const auto& export_ : _glue->exportedIDs() ) {
+        auto spicy_id = std::get<0>(export_);
+        auto zeek_id = std::get<1>(export_);
+
```
bbannier, line 354: We don't need copies for `spicy_id` and `zeek_id` here, let's either use `const auto&` here or a structured binding in the loop header:

```suggestion
    for ( const auto& [spicy_id, zeek_id, _] : _glue->exportedIDs() ) {
```

src/compiler/glue-compiler.cc
```diff
@@ -869,21 +876,19 @@ bool GlueCompiler::compile() {
...
+        if ( auto type = createZeekType(t.first.type, t.second) )
+            preinit_body.addCall("zeek_rt::register_type",
+                                 {builder::string(t.second.namespace_()), builder::string(t.second.local()), *type});
+        else
+            hilti::logger().error(hilti::util::fmt("cannot export Spicy type '%s': %s", t.second, type.error()),
+                                  t.first.location);
 
-        preinit_body.addCall("zeek_rt::register_enum_type", {builder::string(e.id.namespace_()),
-                                                             builder::string(e.id.local()), builder::vector(labels)});
+        exported_type_seen.insert(t.first.id);
```
bbannier, line 889: Nit: Let's get rid of the uses of `first` and `second` while we are at it:

```cpp
    for ( const auto& [tinfo, id] : _driver->exportedTypes() ) {
        if ( auto type = createZeekType(tinfo.type, id) )
            preinit_body.addCall("zeek_rt::register_type",
                                 {builder::string(id.namespace_()), builder::string(id.local()), *type});
        else
            hilti::logger().error(hilti::util::fmt("cannot export Spicy type '%s': %s", id, type.error()),
                                  tinfo.location);

        exported_type_seen.insert(tinfo.id);
```

tests/zeek/export-types.zeek
```diff
@@ -0,0 +1,99 @@
...
+};
+
+# @TEST-END-FILE
+
+# @TEST-START-FILE export.evt
+
+export Test::type_enum;
+export Test::type_record_s as XYZ::type_record_sss;
+export Test::type_record_u;
+export Test::type_record_u2 as type_record_u2;
```
bbannier, line 97: Could you add a test case for exporting into `GLOBAL`?

bbannier (approved):

tests/zeek/export-types.zeek
rsmmr: There is one: `type_record_u2` is mapped into the `GLOBAL` space (`GLOBAL` is just a Zeek-side hack for "no namespace").

bbannier (approved):

PR #152, opened 2022-12-07
Rewrite CI so we can run against zeek5

bbannier:


.cirrus.yml
```diff
@@ -28,7 +32,7 @@ zkg_ubuntu_task:
     fingerprint_script: echo $CIRRUS_TASK_NAME-$CIRRUS_OS
 
   install_spicy_script:
-    - curl -o spicy-dev.deb https://api.cirrus-ci.com/v1/artifact/github/zeek/spicy/docker_ubuntu20/packages/spicy-dev.deb
+    - curl -o spicy-dev.deb https://api.cirrus-ci.com/v1/artifact/github/zeek/spicy/docker_debian11/packages/spicy-dev.deb
```
rsmmr, line 35: Aus Neugier: warum der Wechsel zu Debian?

ci/Dockerfile
```diff
@@ -1,59 +1,49 @@
-# This file is adapted from Spicy upstream's `docker/Dockerfile.ubuntu-20`.
-# Fixes here should probably also happen in that file.
+FROM debian:bullseye-slim AS build
 
-FROM ubuntu:focal
-
-# Increase this to force a rebuild of the Docker image (in Cirrus, in particular).
-ENV IMAGE_VERSION=1
```
rsmmr, line 7: Absicht, dass das raus ist?

rsmmr (approved):

.cirrus.yml
bbannier: I wanted to stay somewhat close to the `Dockerfile` used to publish the Zeek image which uses Debian. Beyond that I don't have a strong preference for what to use here.

ci/Dockerfile
bbannier: I removed this to stay relatively close to the `Dockerfile` used in zeek/zeek. I believe we added this as a workaround for incorrect image caching in Cirrus (which I suspect might be fixed since I haven't seen it for some time), but we could always add a similar line if we want to force a rebuild (this would need a new commit in any case).

PR #156, opened 2022-12-19
Adjust plugin for changes from zeek/spicy#1326.

bbannier:
To address zeek/spicy#1326 we changed the code generation for enums to emit a proper runtime type, in particular a type which has defined semantics for value initialization.

This patch adds compatibility code so the plugin can deal with these generated enums types in addition to the previous format for backwards compatibility.


bbannier:
The matching Spicy change is zeek/spicy#1335.

rsmmr (approved):

PR #158, opened 2023-01-05
Remove support for zeek-4.x.

### LINKED ISSUE #153: Remove support for Zeek 4.x.
I'd say time has come to limit the plugin to Zeek 5.x.

bbannier:
Closes #153.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -15,12 +15,6 @@ zkg_task:
     dockerfile: ci/Dockerfile
     matrix:
       - docker_arguments:
-        - ZEEK_VERSION: "v4.0.3"
-        - ZEEK_CONFFLAGS: ""
-      - docker_arguments:
-        - ZEEK_VERSION: "v4.2.1"
-        - ZEEK_CONFFLAGS: ""
-      - docker_arguments:
         - ZEEK_VERSION: "v5.0.4"
       - docker_arguments:
         - ZEEK_VERSION: "v5.1.1"
--- a/include/zeek-compat.h
+++ b/include/zeek-compat.h
@@ -24,26 +24,12 @@
 #endif
 #endif
 
-#if ZEEK_VERSION_NUMBER < 40000
-#error "Zeek version must be >= 4.0"
-#endif
-
 //// Collect all the Zeek includes here that we need anywhere in the plugin.
 
 #if ZEEK_DEBUG_BUILD
 #ifndef DEBUG
 #define DEBUG
 #endif
-#endif
-
-#include <zeek/packet_analysis/Analyzer.h>
-
-#if ZEEK_VERSION_NUMBER >= 40200
-#include <zeek/Tag.h>
-#else
-#include <zeek/analyzer/Tag.h>
-#include <zeek/file_analysis/Tag.h>
-#include <zeek/packet_analysis/Tag.h>
 #endif
 
 #include <zeek/Conn.h>
@@ -55,6 +41,7 @@
 #include <zeek/Expr.h>
 #include <zeek/IPAddr.h>
 #include <zeek/Reporter.h>
+#include <zeek/Tag.h>
 #include <zeek/Type.h>
 #include <zeek/Val.h>
 #include <zeek/Var.h>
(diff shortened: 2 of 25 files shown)


.cirrus.yml
```diff
@@ -14,12 +14,6 @@ zkg_task:
   container:
     dockerfile: ci/Dockerfile
     matrix:
-      - docker_arguments:
-        - ZEEK_VERSION: "v4.0.3"
-        - ZEEK_CONFFLAGS: ""
-      - docker_arguments:
-        - ZEEK_VERSION: "v4.2.1"
-        - ZEEK_CONFFLAGS: ""
```
rsmmr, line 22: Let's add the most recent 5.x version.

src/plugin.cc
```diff
@@ -247,7 +247,7 @@ void plugin::Zeek_Spicy::Plugin::registerFileAnalyzer(const std::string& name,
     // this point already, so ours won't get initialized anymore.
     c->Initialize();
 
-    trackComponent(c, c->Tag().Type()); // must come after Initialize(); needed for Zeek < 4.2
```
rsmmr, line 250: Not sure what this was about, but the comment seems to say something's not needed anymore now?

src/plugin.cc
```diff
@@ -299,7 +299,7 @@ void plugin::Zeek_Spicy::Plugin::registerPacketAnalyzer(const std::string& name,
...
+    trackComponent(c, c->Tag().Type()); // Must come after Initialize().
```
rsmmr, line 302: Same.

include/zeek-compat.h
```diff
@@ -55,24 +41,21 @@
...
 #include <zeek/analyzer/Analyzer.h>
 #include <zeek/analyzer/Manager.h>
 #include <zeek/analyzer/protocol/pia/PIA.h>
 #include <zeek/analyzer/protocol/tcp/TCP.h>
-#if ZEEK_VERSION_NUMBER < 40100 // Zeek < 4.1
-#include <zeek/Sessions.h>
-#include <zeek/analyzer/protocol/udp/UDP.h>
-#else // Zeek >= 4.1
-#include <zeek/session/Manager.h>
-#endif
```
rsmmr, line 70: From the diff it's hard to say if some of the `#else` blocks being removed were triggering for Zeek >= 5.0. Please just double-check that nothing goes out accidentally (although then things would presumably break, so probably fine as it is).

rsmmr (approved):
One more thing to check, but hard to see from the diff: If there are now cases in `zeek::compat` where we don't have to version-switch at all anymore, just update the calling code to remove the usage of the `compat` wrapper altogether.

include/zeek-compat.h
bbannier: This should be fine, the include for `zeek/session/Manager.h` stays, the one for `zeek/Sessions.h` is removed. What makes these include changes harder to review is that they also end up resorted (previously the preprocessor calls acted as barriers).

src/plugin.cc
bbannier: These `trackComponent` calls update `Plugin::_analyzer_name_to_tag_type` which we seem to be using to update linker scopes for pre-registered analyzers, see e.g., https://github.com/zeek/spicy-plugin/blob/09184f9f707b275c90da50a07ca7466dfc7ecd4c/src/plugin.cc#L163-L178. Not sure, but maybe we'll be able to remove this once we remove plugin JIT mode.

.cirrus.yml
bbannier: Not sure I am missing something, but just below we run against `v5.1.1` which should be the latest release.

src/plugin.cc
bbannier: See above.

bbannier:
> One more thing to check, but hard to see from the diff: If there are now cases in `zeek::compat` where we don't have to version-switch at all anymore, just update the calling code to remove the usage of the `compat` wrapper altogether.

I went through the namespace and nothing in it remains. I added two new patches to the PR cleaning it up, PTAL.

.cirrus.yml
rsmmr: I'm just blind. :-)

rsmmr (approved):
Great to see the new, tiny `zeek-compat.h`!

PR #163, opened 2023-01-21
Revert moving Zeek includes out of `zeek-compat.h.`

rsmmr:
This reverts most of `58facecloth`. The problem is that the Zeek includes need some consistent handling in terms of preprocessor setup (`DEBUG`) and inclusion of `zeek-config.h`. The old `zeek-compat.h` used to handle that, but we lost it by moving the includes over into individual files. As a result, I saw some ill-defined crashes inside the plugin, depending on local configuration/environment.

This reverts most of the previous the change because it's easier/safer to handle this centrally than adding the logic to all files needing to include some Zeek header.

### DIFF
--- a/include/zeek-compat.h
+++ b/include/zeek-compat.h
@@ -6,7 +6,7 @@
 
 #include <zeek-spicy/autogen/config.h>
 
-#include <zeek/util.h>
+#include <zeek/zeek-config.h>
 
 #ifdef ZEEK_VERSION_NUMBER
 #if ZEEK_SPICY_VERSION_NUMBER != ZEEK_VERSION_NUMBER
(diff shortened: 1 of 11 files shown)


include/zeek-compat.h
```diff
@@ -19,7 +19,42 @@
 #endif
 #endif
 
+//// Collect all the Zeek includes here that we need anywhere in the plugin.
+
+#if ZEEK_DEBUG_BUILD
+#ifndef DEBUG
+#define DEBUG
```
bbannier, line 26: Let's remember here whether we forced `DEBUG` and only `undef` it below if we did, e.g.,

```suggestion
#define SPICY_PLUGIN_DEBUG_DEFINED
#define DEBUG
```

include/zeek-compat.h
```diff
@@ -19,7 +19,42 @@
...
+#include <zeek/analyzer/protocol/pia/PIA.h>
+#include <zeek/analyzer/protocol/tcp/TCP.h>
+#include <zeek/file_analysis/Analyzer.h>
+#include <zeek/file_analysis/File.h>
+#include <zeek/file_analysis/Manager.h>
+#include <zeek/module_util.h>
+#include <zeek/packet_analysis/Analyzer.h>
+#include <zeek/plugin/Plugin.h>
+
+#undef DEBUG
```
bbannier, line 54: With above flag this would become
```suggestion
#ifdef SPICY_PLUGIN_DEBUG_DEFINED
#undef DEBUG
#endif
```

bbannier (commented):
Thanks for the cleanup, I somehow didn't hit this error. This is still better than the state before 5313a1bf79d03c3170c69c7a27c815be104b2cda where would inconsistently include Zeek headers in other files which is now gone.

Could you fix the commit message to reference the commit 5313a1bf79d03c3170c69c7a27c815be104b2cda where the removal was merged?

bbannier (dismissed):

Issue #154, opened 2022-12-13
Remove JIT mode


rsmmr:
Per discussion in https://community.zeek.org/t/rfc-removing-jit-mode-from-spicys-zeek-plugin/6854. I haven't heard any objections to this proposal.

PR #164, opened 2023-01-23
Remove JIT support from plugin

rsmmr:
Per #154 .


src/plugin/plugin.cc
```diff
@@ -800,7 +698,10 @@ int plugin::Zeek_Spicy::Plugin::HookLoadFile(const LoadType type, const std::str
     }
 
     if ( ext == ".spicy" || ext == ".evt" || ext == ".hlt" )
-        reporter::fatalError(hilti::rt::fmt("cannot load '%s', Spicy plugin was not compiled with JIT support", file));
+        reporter::fatalError(
+            hilti::rt::fmt("cannot load '%s': Spicy plugin no longer supports just-in-time compilation; use 'spicyz' "
+                           "to precompile analyzers",
```
bbannier, line 703: This is a nice error message the moment a user bumps versions, but immediately after that becomes dated. Maybe instead

```suggestion
        reporter::fatalError(hilti::rt::fmt("cannot load '%s': analyzers need to be precompiled with 'spicyz' ", file));
```

bbannier (dismissed):
Another nice cleanup :)

PR #165, opened 2023-01-26
Adapt to Spicy's changes to exception hierarchy.

### LINKED ISSUE #160: Treat integer overflows as analyzer violations (question/proposal)
After discussing with @rsmmr: 

Should integer overflow/underflow errors trigger implicit analyzer violations rather than analyzer errors? I'm not sure what's right or nicer, but currently an integer overflow is treated differently than, for example, unexpected input data or regex mismatch.

As of today, Spicy parser writers need  diligently protect subtraction/addition of parsed values with `&requires` fields to not trigger analyzer errors. I believe this topic came up with an analyzer from @keithjjones before and he may have stumbled over it enough to have an opinion.

The following grammar assumes length is at least `9` due to how length, major, minor and the terminating `\x00` are all supposed to be included in length by specification (this is PostgreSQL). If they are not, that's a violation of the protocol.

```
module Mini;

public type FrontendMessage = unit {
    length:     uint32;
    major:      uint16;
    minor:      uint16;
    # Capture everything but length, major, minor and "\x00"
    parameters: bytes &size=self.length - 9;
    :           b"\x00";
};
```
The following pcap contains as a first message such a FrontendMessage, but with the length set to `3` instead of the expected larger value: [bad-startup-message-1.pcap.zip](https://github.com/zeek/spicy-plugin/files/10374457/bad-startup-message-1.pcap.zip)

Running this through Zeek with an .evt file for port 5432, causes the following reporter.log entry because `zeek::reporter->AnalyzerError()` is invoked. By default, this also makes it to stderr:
```
1673270524.038183       Reporter::ERROR integer overflow        /home/awelzel/corelight-oss/projects/spicy-mini/analyzer/mini.spicy, line 8
```
Further, without an explicit `FrontendMessage%error` handler calling `zeek::reject_protocol()`, there is no `analyzer_violation` event triggered. The integer overflow can be prevented by putting a `&requires=(self.length >= 9)` which then makes things nicer causing an analyzer violation:
```
violation, Analyzer::ANALYZER_SPICY_MINI, C21rnr2KzAmlQ38u52, &requires failed (/home/awelzel/corelight-oss/projects/spicy-mini/analyzer/mini.spicy:4:24)
```

Adapting the grammar as follows to expect `\x12\x34` as major version to cause a different violation: 
```
- major:      uint16;
+ major:      b"\x12\x34";
```
This triggers an analyzer violation and there's no entry in reporter.log created. The info below is from the `analyzer_violation` event (`\x34` is ascii `4`):
```
violation, Analyzer::ANALYZER_SPICY_MINI, CzvCrF1Hevsxs7VIzl, expecting '\x124' (/home/awelzel/corelight-oss/projects/spicy-mini/analyzer/mini.spicy:5:17)
```

---

.evt file:
```
import Mini;
import Zeek_Mini;

protocol analyzer spicy::Mini over TCP:
    parse originator with Mini::FrontendMessage,
    parse responder with Mini::BackendMessage,
    port 5432/tcp;

on Mini::FrontendMessage -> event Mini::frontend_message($conn, $is_orig, self.parameters);
```

analyzer violation handler:
```
event analyzer_violation_info(atype: AllAnalyzers::Tag, info: AnalyzerViolationInfo) {
        print "violation", atype, info$c$uid, info$reason;
}
```

rsmmr:
We now consistently report runtime errors (including actual parsing
errors) as analyzer violations; and o longer as analyzer errors. The
former go into DPD and `analyzer.log`, while the latter would end up
in `reporter.log` (which isn't really helpful).

The commit also takes the opportunity to report file analyzer
violations through Zeek's new API for that when available; and passes
offending data along with the violation for logging on the Zeek side.

This goes with https://github.com/zeek/spicy/pull/1362. It remains
working with older Spicy versions, too, but for them the output may
change (which is why we disable one test for older version). With that
branch in use, the output is what we want.

Closes #160.


rsmmr:
This won't pass CI before topic/robin/cleanup-exceptions gets merged into `main`, but I'll turn it read for review now.

src/file-analyzer.cc
```diff
@@ -89,17 +89,27 @@ bool FileAnalyzer::Process(int len, const u_char* data) {
...
         hilti::rt::context::CookieSetter _(&_state.cookie());
         _state.process(len, reinterpret_cast<const char*>(data));
-    } catch ( const spicy::rt::ParseError& e ) {
-        STATE_DEBUG_MSG(hilti::rt::fmt("parse error, triggering analyzer violation: %s", e.what()));
+    } catch ( const hilti::rt::RuntimeError& e ) {
+        STATE_DEBUG_MSG(hilti::rt::fmt("error during parsing, triggering analyzer violation: %s", e.what()));
         auto tag = OurPlugin->tagForFileAnalyzer(_state.cookie().analyzer->Tag());
-        // TODO: Report a weird from this file analysis once Zeek supports it.
+#if ZEEK_VERSION_NUMBER >= 50200
+        AnalyzerViolation(e.what(), reinterpret_cast<const char*>(data), len, tag);
```
bbannier, line 109: Nice, hadn't realized this now worked.

include/runtime-support.h
```diff
@@ -75,9 +75,9 @@ class TypeMismatch : public hilti::rt::UserException {
 /**
  * Exception thrown by the runtime library when Zeek has flagged a problem.
  */
-class ZeekError : public hilti::rt::UserException {
+class ZeekError : public spicy::compat::UsageError {
 public:
-    using hilti::rt::UserException::UserException;
+    using spicy::compat::UsageError::UsageError;
```
bbannier, line 80: Ditto.

include/runtime-support.h
```diff
@@ -29,37 +29,37 @@ namespace spicy::zeek::rt {
...
 
 /**
  * Exception thrown by event generation code if functionality is used
  * that the current build does not support.
  */
-class Unsupported : public hilti::rt::UserException {
+class Unsupported : public spicy::compat::UsageError {
 public:
-    using hilti::rt::UserException::UserException;
+    using spicy::compat::UsageError::UsageError;
```
bbannier, line 52: Ditto.

include/runtime-support.h
```diff
@@ -29,37 +29,37 @@ namespace spicy::zeek::rt {
...
 
 /**
  * Exception thrown by event generation code if the values can't be converted
  * to Zeek.
  */
-class InvalidValue : public hilti::rt::UserException {
+class InvalidValue : public spicy::compat::UsageError {
 public:
-    using hilti::rt::UserException::UserException;
+    using spicy::compat::UsageError::UsageError;
```
bbannier, line 43: Ditto.

include/runtime-support.h
```diff
@@ -29,37 +29,37 @@ namespace spicy::zeek::rt {
  * Exception thrown by event generation code if the value of an `$...`
  * expression isn't available.
  */
-class ValueUnavailable : public hilti::rt::UserException {
+class ValueUnavailable : public spicy::compat::UsageError {
 public:
-    using hilti::rt::UserException::UserException;
+    using spicy::compat::UsageError::UsageError;
```
bbannier, line 34: Does this actually work if `UsageError` is actually an `UserException`?

rsmmr: I think it does, but maybe I'm missing something on why it shouldn't? Haven't seen any problems during testing yet. Will leave as is and see if something shows up.

PR #166, opened 2023-01-26
Switch to new API for accessing the HLTO scope.

rsmmr:
This goes with https://github.com/zeek/spicy/pull/1358, but remains
working with older Spicy versions.

### DIFF
--- a/src/compiler/glue-compiler.cc
+++ b/src/compiler/glue-compiler.cc
@@ -22,6 +22,12 @@ using namespace spicy::zeek;
 
 namespace builder = hilti::builder;
 
+#if SPICY_VERSION_NUMBER >= 10700
+static inline auto _linker_scope() { return hilti::builder::scope(); }
+#else
+static inline auto _linker_scope() { return hilti::builder::call("hilti::linker_scope", {}); }
+#endif
+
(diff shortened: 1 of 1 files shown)


rsmmr:
This won't pass CI before https://github.com/zeek/spicy/pull/1358 gets merged into `main`, but I'll turn it read for review now.

src/compiler/glue-compiler.cc
```diff
@@ -22,6 +22,12 @@ using namespace spicy::zeek;
 
 namespace builder = hilti::builder;
 
+#if SPICY_VERSION_NUMBER >= 10700
+static inline auto _linker_scope() { return hilti::builder::scope(); }
+#else
+static inline auto _linker_scope() { return hilti::builder::call("hilti::linker_scope", {}); }
+#endif
+
```
bbannier, line 30: If #165 you introduced `spicy_compat.h`. It might make sense to add these functions there instead as well.

rsmmr: Actually, #164 gets rid of `spicy_compat.h` once more, because of the new separation there. That's why this is all a bit inconsistent across branches, I'll unify that before the JIT removal lands.

PR #167, opened 2023-01-26
Add a script-level constant indicating Spicy availability.

rsmmr:
This enables Zeek scripts to test for availability of Spicy through
`@ifdef ( Spicy::available )`.

### DIFF
--- a/scripts/__preload__.zeek
+++ b/scripts/__preload__.zeek
@@ -4,6 +4,9 @@ module Spicy;
 
 export {
 # doc-options-start
+    # Constant for testing if Spicy is available.
+    const available = T;
+
     ## Activate compile-time debugging output for given debug streams (comma-separated list).
     const codegen_debug = "" &redef;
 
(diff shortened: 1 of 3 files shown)


scripts/__preload__.zeek
```diff
@@ -4,6 +4,9 @@ module Spicy;
 
 export {
 # doc-options-start
+    # Constant for testing if Spicy is available.
+    const available = T;
+
```
bbannier, line 9: It looks like this won't be available in Zeek bare mode. Could we move this to `bare.zeek` (unsure how that interacts with `consts.bif` currently).

rsmmr: Good observation—but it's actually working as expected: in bare mode, plugins aren't active by default, and hence Spicy will indeed not be available. Activation needs to be explicit in bare mode: `zeek -b Zeek::Spicy ...`. I just confirmed that `available` then works correctly. I have also added a test now that checks the 3 cases that we have here.

PR #168, opened 2023-01-26
Add BTest alternative for running tests when plugin is built into Zeek.

rsmmr:


bbannier (approved):
Slightly related, we were recently discussing whether we should have Zeek execute Spicy's BTest suite once in a while as well; something like this in Spicy would allow that.

PR #175, opened 2023-01-30
Remove more code that we no longer needs without JIT.

rsmmr:


bbannier (dismissed):

src/compiler/driver.cc
```diff
@@ -138,138 +138,6 @@ Driver::Driver(std::unique_ptr<GlueCompiler> glue, const char* argv0, hilti::rt:
 
 Driver::~Driver() {}
 
-void Driver::usage(std::ostream& out) {
```
bbannier, line 141: Let's also remove the declaration.

src/compiler/driver.cc
```diff
@@ -138,138 +138,6 @@ Driver::Driver(std::unique_ptr<GlueCompiler> glue, const char* argv0, hilti::rt:
...
-           "  -C             Dump all generated code to disk for debugging.\n"
-           "  -D <streams>   Activate compile-time debugging output for given debug streams (comma-separated).\n"
-           "  -L <path>      Add path to list of directories to search when importing modules.\n"
-           "  -R             Report a break-down of compiler's execution time.\n"
-           "  -V             Don't validate ASTs (for debugging only).\n"
-           "  -X <addl>      Implies -d and adds selected additional instrumentation (comma-separated).\n"
-           "\n";
-}
-
-hilti::Result<hilti::Nothing> Driver::parseOptionsPreScript(const std::string& options) {
```
bbannier, line 157: Ditto.

PR #176, opened 2023-01-30
Make it possible to test against a certain Spicy branch.

bbannier:


rsmmr (approved):
I trust you that that works. :-)

PR #178, opened 2023-02-03
Support hook arguments in EVT files.

### LINKED ISSUE #162: Consider making it possible passing error message in EVT hook mapping
Spicy allows accessing error messages in `%error` hooks. There currently seems no way to directly pass the message into Zeek events in the EVT mapping. We should consider extending the syntax to make this possible.

rsmmr:
Some built-in unit hooks allow to receive parameters. To enable access
to them inside EVT files, we now support the same syntax there as with
`on ...` declarations in Spicy source code. For example, to gain
access to the error message with `%error`:

    on SSH::Banner::%error(msg: string) -> event MyScript::my_event(msg);

Closes #162.


rsmmr:
`static_zeek_master` failure is the same as in `main`.

src/compiler/glue-compiler.cc
```diff
@@ -101,6 +101,36 @@ static hilti::ID extract_id(const std::string& chunk, size_t* i) {
...
+
+    auto id = extract_id(chunk, i);
+
+    if ( ! looking_at(chunk, *i, ":") )
+        throw ParseError("expected ':'");
+
+    eat_token(chunk, i, ":");
+
+    auto type = extract_type(chunk, i);
+    return builder::parameter(id, type);
```
bbannier, line 131: Nit: We could move the parameters here,

```suggestion
    return builder::parameter(std::move(id), std::move(type));
```

bbannier (approved):

PR #179, opened 2023-02-03
Allow duplicate Zeek-side type registration as long as types match.

### LINKED ISSUE #177: Automatic Zeek type creation clashes with using common base libraries
Probably unsurprising, if one adds the same Spicy source file to multiple HLTO files, one cannot use these HLTO files at the same time since the same types on the Zeek side are declared multiple times (from multiple HLTO files). This is bad since it makes it hard to reuse identical helper types and functions in multiple modules (e.g., imagine a world where multiple independent parsers pull in ASN1 parsers for subparser).

At least for my tests this only seems to happen for modules directly added to the `spicyz` invocation like one would do for proper dependency tracking in e.g., a CMake setup; if one instead simply imports and uses the module there seems to be no issue.

We should figure out why there is a mismatch between simply imported and explicitly added modules, and also check whether we need to reject all such duplicate definitions (tricky bit: even if only looking at types generated from the Spicy side, it probably is hard to decide whether identical names indeed refer to identical types).

Reproducer in the Zeek repo (e.g., `testing/btest/spicy/spicyz-common.spicy`), zeek/zeek@39f0b780436cf711c8d084bd486cf188e259e892:

```ruby
# Works.
# @TEST-EXEC: ${BUILD}/src/builtin-plugins/spicy-plugin/bin/spicyz -o a.hlto a.spicy
# @TEST-EXEC: ${BUILD}/src/builtin-plugins/spicy-plugin/bin/spicyz -o b.hlto b.spicy
# @TEST-EXEC: zeek a.hlto b.hlto

# Breaks.
# @TEST-EXEC: ${BUILD}/src/builtin-plugins/spicy-plugin/bin/spicyz -o a.hlto a.spicy common.spicy
# @TEST-EXEC: ${BUILD}/src/builtin-plugins/spicy-plugin/bin/spicyz -o b.hlto b.spicy common.spicy
# @TEST-EXEC: zeek a.hlto b.hlto

# @TEST-START-FILE common.spicy
module common;

public type E = enum {
	X,
	Y,
};
# @TEST-END-FILE

# @TEST-START-FILE a.spicy
module a;

import common;
global x: common::E;
# @TEST-END-FILE

# @TEST-START-FILE b.spicy
module b;

import common;
global x: common::E;
# @TEST-END-FILE
```

```console
$ btest -dv picy/spicyz-common.spicy
spicy.spicyz-common ...
  > ${BUILD}/src/builtin-plugins/spicy-plugin/bin/spicyz -o a.hlto a.spicy
  > ${BUILD}/src/builtin-plugins/spicy-plugin/bin/spicyz -o b.hlto b.spicy
  > zeek a.hlto b.hlto
  > ${BUILD}/src/builtin-plugins/spicy-plugin/bin/spicyz -o a.hlto a.spicy common.spicy
  > ${BUILD}/src/builtin-plugins/spicy-plugin/bin/spicyz -o b.hlto b.spicy common.spicy
  > zeek a.hlto b.hlto
... spicy.spicyz-common failed
  % 'zeek a.hlto b.hlto' failed unexpectedly (exit code 1)
  % cat .stderr
  error: attempt to overwrite already defined Zeek type 'common::E'

1 of 1 test failed
```

rsmmr:
Closes #177.

### DIFF
--- a/src/plugin/plugin.cc
+++ b/src/plugin/plugin.cc
@@ -238,10 +238,20 @@ void plugin::Zeek_Spicy::Plugin::registerPacketAnalyzer(const std::string& name,
 void plugin::Zeek_Spicy::Plugin::registerType(const std::string& id, const ::zeek::TypePtr& type) {
     auto [ns, local] = parseID(id);
 
-    if ( ::zeek::detail::lookup_ID(local.c_str(), ns.c_str()) ) {
-        // Note, this is unlikely  to trigger, because we run before Zeek
-        // registers its own IDs (but then Zeek will catch it).
-        reporter::error(hilti::rt::fmt("attempt to overwrite already defined Zeek type '%s'", id));
+    if ( auto old = ::zeek::detail::lookup_ID(local.c_str(), ns.c_str()) ) {
+        // This is most likely to trigger for IDs that other Spicy modules
+        // register. If we two Spicy modules need the same type, that's ok as
+        // long as they match.
+        if ( ! old->IsType() ) {
+            reporter::error(hilti::rt::fmt("Zeek type registration failed for '%s': ID already exists, but is not a type", id));
+            return;
+        }
+
+        if ( ! zeek::same_type(type, old->GetType()) ) {
+            reporter::error(hilti::rt::fmt("Zeek type registration failed for '%s': Type already exists, but differs", id));
+        }
+
+        ZEEK_DEBUG(hilti::rt::fmt("Not re-registering Zeek type %s: identical type already exists", id));
         return;
     }
 
--- a/tests/zeek/double-type-registration.zeek
+++ b/tests/zeek/double-type-registration.zeek
@@ -1,0 +1,32 @@
+# @TEST-EXEC: spicyz -L . -o a.hlto a.spicy
+# @TEST-EXEC: spicyz -L . -o b.hlto b.spicy
+# @TEST-EXEC: zeek a.hlto b.hlto
+#
+# @TEST-EXEC: spicyz -o a.hlto a.spicy common.spicy
+# @TEST-EXEC: spicyz -o b.hlto b.spicy common.spicy
+# @TEST-EXEC: zeek a.hlto b.hlto
+#
+# @TEST-DOC: Regression test for #177.
+
+# @TEST-START-FILE common.spicy
+module common;
+
+public type E = enum {
+	X,
+	Y,
+};
+# @TEST-END-FILE
+
+# @TEST-START-FILE a.spicy
+module a;
+
+import common;
+global x: common::E;
+# @TEST-END-FILE
+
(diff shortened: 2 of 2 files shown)


tests/zeek/double-type-registration.zeek
```diff
@@ -0,0 +1,32 @@
+# @TEST-EXEC: spicyz -L . -o a.hlto a.spicy
+# @TEST-EXEC: spicyz -L . -o b.hlto b.spicy
```
bbannier, line 2: Let's drop the `-L .` here like we do below. It should make no difference, but made me wonder whether it does.

src/plugin/plugin.cc
```diff
@@ -238,10 +238,20 @@ void plugin::Zeek_Spicy::Plugin::registerPacketAnalyzer(const std::string& name,
...
+    if ( auto old = ::zeek::detail::lookup_ID(local.c_str(), ns.c_str()) ) {
+        // This is most likely to trigger for IDs that other Spicy modules
+        // register. If we two Spicy modules need the same type, that's ok as
+        // long as they match.
+        if ( ! old->IsType() ) {
+            reporter::error(hilti::rt::fmt("Zeek type registration failed for '%s': ID already exists, but is not a type", id));
+            return;
+        }
+
+        if ( ! zeek::same_type(type, old->GetType()) ) {
```
bbannier, line 250: Ah, I wasn't aware of `zeek::same_type`, exactly what we needed.

src/plugin/plugin.cc
```diff
@@ -238,10 +238,20 @@ void plugin::Zeek_Spicy::Plugin::registerPacketAnalyzer(const std::string& name,
 void plugin::Zeek_Spicy::Plugin::registerType(const std::string& id, const ::zeek::TypePtr& type) {
     auto [ns, local] = parseID(id);
 
-    if ( ::zeek::detail::lookup_ID(local.c_str(), ns.c_str()) ) {
-        // Note, this is unlikely  to trigger, because we run before Zeek
-        // registers its own IDs (but then Zeek will catch it).
-        reporter::error(hilti::rt::fmt("attempt to overwrite already defined Zeek type '%s'", id));
+    if ( auto old = ::zeek::detail::lookup_ID(local.c_str(), ns.c_str()) ) {
```
bbannier, line 241: Style nit: No need to make a copy of the `IDPtr` here.
```suggestion
    if ( const auto& old = ::zeek::detail::lookup_ID(local.c_str(), ns.c_str()) ) {
```

bbannier (approved):

tests/zeek/double-type-registration.zeek
rsmmr: I needed it, otherwise it didn't compile.

PR #182, opened 2023-02-07
Allow more file system operations to fail.

### LINKED ISSUE #181: Unhandled exception in searchModules (was: Crash on startup on FreeBSD 12.4-RELEASE)
invoking zeek via `/usr/local/etc/rc.d/zeek start` or `/usr/local/etc/rc.d/zeek deploy` causes a core dump.

```
Zeek 5.0.5   
FreeBSD 12.4-RELEASE                                                                                                                

Zeek plugins: (none found)
                                                                                                                                    
Core file: zeek.core                                                                                                                [New LWP 100234]                                                                                                                    
Core was generated by `/usr/local/bin/zeek -N'.
Program terminated with signal SIGABRT, Aborted.
Sent by thr_kill() from pid 63881 and user 782.
#0  0x00000008031f0b8a in thr_kill () from /lib/libc.so.7

Thread 1 (LWP 100234):
#0  0x00000008031f0b8a in thr_kill () from /lib/libc.so.7
zeek/zeek#1  0x00000008031eef54 in raise () from /lib/libc.so.7
zeek/zeek#2  0x0000000803167449 in abort () from /lib/libc.so.7
zeek/zeek#3  0x0000000802fe1259 in ?? () from /lib/libcxxrt.so.1
zeek/zeek#4  0x0000000001c227ad in ghc::filesystem::recursive_directory_iterator::operator++() ()                                           
zeek/zeek#5  0x0000000001c20785 in plugin::Zeek_Spicy::Plugin::searchModules(std::__1::basic_string<char, std::__1::char_traits<char>, std::_
_1::allocator<char> > const&) ()                                                                                                    
zeek/zeek#6  0x0000000001c1d108 in plugin::Zeek_Spicy::Plugin::autoDiscoverModules() ()                                                     
zeek/zeek#7  0x0000000001c1ce29 in plugin::Zeek_Spicy::Plugin::InitPreScript() ()
zeek/zeek#8  0x0000000001605630 in zeek::plugin::Manager::InitPreScript() ()
zeek/zeek#9  0x0000000001459761 in zeek::detail::setup(int, char**, zeek::Options*) ()
zeek/zeek#10 0x0000000001c981d3 in main () 

==== No reporter.log

==== stderr.log
/usr/local/share/zeekctl/scripts/run-zeek: line 53: ulimit: data seg size: cannot modify limit: Operation not permitted
/usr/local/share/zeekctl/scripts/run-zeek: line 61: ulimit: core file size: cannot modify limit: Operation not permitted
/usr/local/share/zeekctl/scripts/run-zeek: line 110: 16734 Abort trap              (core dumped) nohup "$myzeek" "$@"

==== stdout.log
max memory size             (kbytes, -m) unlimited
data seg size               (kbytes, -d) 33554432
virtual memory              (kbytes, -v) unlimited
core file size              (blocks, -c) 81920000

==== .cmdline
-i vlan2 -U .status -p zeekctl -p zeekctl-live -p standalone -p local -p zeek local.zeek zeekctl zeekctl/standalone zeekctl/auto

==== .env_vars
PATH=/usr/local/bin:/usr/local/share/zeekctl/scripts:/home/nf/bin:/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/bin:/sbin
ZEEKPATH=/usr/local/spool/installed-scripts-do-not-touch/site::/usr/local/spool/installed-scripts-do-not-touch/auto:/usr/local/share
/zeek:/usr/local/share/zeek/policy:/usr/local/share/zeek/site:/usr/local/share/zeek/builtin-plugins
CLUSTER_NODE=

==== .status
INITIALIZING [main]

==== No prof.log

==== No packet_filter.log

==== No loaded_scripts.log
```


my local.zeek:

```
##! Local site policy. Customize as appropriate.      
##!                                                               
##! This file will not be overwritten when upgrading or reinstalling!
                                 
# Installation-wide salt value that is used in some digest hashes, e.g., for
# the creation of file IDs. Please change this to a hard to guess value.
redef digest_salt = "REDACTED";
                                                                                                                                    
# This script logs which scripts were loaded during each run.
@load misc/loaded-scripts                                         

# Apply the default tuning scripts for common tuning settings.                                                                      
@load tuning/defaults                                                                                                               
                                                                                                                                    
# Estimate and log capture loss.                                  
@load misc/capture-loss                                           
                                                                  
# Enable logging of memory, packet and lag statistics.                                                                              
@load misc/stats                                                  
                                                                  
# Load the scan detection script.  It's disabled by default because
# it often causes performance issues.                                                                                               
#@load misc/scan                                                  
                                                                  
# Detect traceroute being run on the network. This could possibly cause
# performance trouble when there are a lot of traceroutes on your network.                                                          
# Enable cautiously.                                                                                                                
#@load misc/detect-traceroute                                                                                                       
                                                                  
# Generate notices when vulnerable versions of software are discovered.
# The default is to only monitor software found in the address space defined      
# as "local".  Refer to the software framework's documentation for more
# information.              
@load frameworks/software/vulnerable                              
                                                                                                                                    
# Detect software changing (e.g. attacker installing hacked SSHD).                                                                  
@load frameworks/software/version-changes                                                                                           
                                                                                                                                    
# This adds signatures to detect cleartext forward and reverse windows shells.
@load-sigs frameworks/signatures/detect-windows-shells                                                                              
                                                                                                                                    
# Load all of the scripts that detect software in various protocols.                                                                
@load protocols/ftp/software
@load protocols/smtp/software
@load protocols/ssh/software
@load protocols/http/software
# The detect-webapps script could possibly cause performance trouble when
# running on live traffic.  Enable it cautiously.
#@load protocols/http/detect-webapps

# This script detects DNS results pointing toward your Site::local_nets
# where the name is not part of your local DNS zone and is being hosted
# externally.  Requires that the Site::local_zones variable is defined.
@load protocols/dns/detect-external-names

# Script to detect various activity in FTP sessions.
@load protocols/ftp/detect

# Scripts that do asset tracking. 
@load protocols/conn/known-hosts
@load protocols/conn/known-services
@load protocols/ssl/known-certs

# This script enables SSL/TLS certificate validation.
@load protocols/ssl/validate-certs

# This script prevents the logging of SSL CA certificates in x509.log
@load protocols/ssl/log-hostcerts-only

# Uncomment the following line to check each SSL certificate hash against the ICSI
# certificate notary service; see http://notary.icsi.berkeley.edu .
# @load protocols/ssl/notary

# If you have GeoIP support built in, do some geographic detections and
# logging for SSH traffic.
@load protocols/ssh/geo-data
# Detect hosts doing SSH bruteforce attacks.
@load protocols/ssh/detect-bruteforcing
# Detect logins using "interesting" hostnames.
@load protocols/ssh/interesting-hostnames

# Detect SQL injection attacks.
@load protocols/http/detect-sqli

#### Network File Handling ####

# Enable MD5 and SHA1 hashing for all files.
@load frameworks/files/hash-all-files

# Detect SHA1 sums in Team Cymru's Malware Hash Registry.
@load frameworks/files/detect-MHR 

# Extend email alerting to include hostnames
@load policy/frameworks/notice/extend-email/hostnames

# Uncomment the following line to enable detection of the heartbleed attack. Enabling
# this might impact performance a bit.
# @load policy/protocols/ssl/heartbleed

# Uncomment the following line to enable logging of connection VLANs. Enabling
# this adds two VLAN fields to the conn.log file.
# @load policy/protocols/conn/vlan-logging

# Uncomment the following line to enable logging of link-layer addresses. Enabling
# this adds the link-layer address for each connection endpoint to the conn.log file.
# @load policy/protocols/conn/mac-logging

# Uncomment this to source zkg's package state
# @load packages

redef restrict_filters += {
    ["broadcast"] = "not host 10.69.67.255"
};
```

bbannier:
We previously would use the HILTI filesystem API in a couple places without taking into account that operation might fail e.g., due to insufficient file system permissions. Since we did not explicitly handle the errors this lead to an exception escape into the calling process (in this case Zeek) which didn't anticipate that either and just aborted.

With this patch we now should probably handle these cases gracefully as well so no more exceptions escape from file system errors.

Closes #181.

### DIFF
--- a/src/compiler/driver.cc
+++ b/src/compiler/driver.cc
@@ -125,12 +125,26 @@ Driver::~Driver() {}
 
 hilti::Result<hilti::Nothing> Driver::loadFile(hilti::rt::filesystem::path file,
                                                const hilti::rt::filesystem::path& relative_to) {
+    std::error_code ec;
     if ( ! relative_to.empty() && file.is_relative() ) {
-        if ( auto p = relative_to / file; hilti::rt::filesystem::exists(p) )
+        auto p = relative_to / file;
+        auto exists = hilti::rt::filesystem::exists(p, ec);
+
+        if ( ec )
+            return hilti::rt::result::Error(
+                hilti::util::fmt("error computing path of %s relative to %s: %s", file, relative_to, ec.message()));
+
+        if ( exists )
             file = p;
     }
 
-    if ( ! hilti::rt::filesystem::exists(file) ) {
+    auto exists = hilti::rt::filesystem::exists(file, ec);
+
+    if ( ec )
+        return hilti::rt::result::Error(
+            hilti::util::fmt("cannot check whether file %s exists: %s", file, ec.message()));
+
+    if ( ! exists ) {
         if ( auto path = hilti::util::findInPaths(file, hiltiOptions().library_paths) )
             file = *path;
         else
--- a/src/plugin/plugin.cc
+++ b/src/plugin/plugin.cc
@@ -11,6 +11,7 @@
 #include <hilti/rt/fmt.h>
 #include <hilti/rt/init.h>
 #include <hilti/rt/library.h>
+#include <hilti/rt/logging.h>
 #include <hilti/rt/types/vector.h>
 #include <hilti/rt/util.h>
 
@@ -243,12 +244,14 @@ void plugin::Zeek_Spicy::Plugin::registerType(const std::string& id, const ::zee
         // register. If we two Spicy modules need the same type, that's ok as
         // long as they match.
         if ( ! old->IsType() ) {
-            reporter::error(hilti::rt::fmt("Zeek type registration failed for '%s': ID already exists, but is not a type", id));
+            reporter::error(
+                hilti::rt::fmt("Zeek type registration failed for '%s': ID already exists, but is not a type", id));
             return;
         }
 
         if ( ! zeek::same_type(type, old->GetType()) ) {
-            reporter::error(hilti::rt::fmt("Zeek type registration failed for '%s': Type already exists, but differs", id));
+            reporter::error(
+                hilti::rt::fmt("Zeek type registration failed for '%s': Type already exists, but differs", id));
         }
 
         ZEEK_DEBUG(hilti::rt::fmt("Not re-registering Zeek type %s: identical type already exists", id));
@@ -726,7 +729,10 @@ void plugin::Zeek_Spicy::Plugin::loadModule(const hilti::rt::filesystem::path& p
     try {
         // If our auto discovery ends up finding the same module multiple times,
         // we ignore subsequent requests.
-        auto canonical_path = hilti::rt::filesystem::canonical(path);
+        std::error_code ec;
+        auto canonical_path = hilti::rt::filesystem::canonical(path, ec);
+        if ( ec )
+            hilti::rt::fatalError(hilti::rt::fmt("could not compute canonical path for %s: %s", path, ec.message()));
 
         if ( auto [library, inserted] = _libraries.insert({canonical_path, hilti::rt::Library(canonical_path)});
              inserted ) {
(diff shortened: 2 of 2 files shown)


src/plugin/plugin.cc
```diff
@@ -768,17 +773,29 @@ void plugin::Zeek_Spicy::Plugin::searchModules(const std::string& paths) {
...
-            if ( e.is_regular_file() && e.path().extension() == ".hlto" )
-                loadModule(e.path());
+        auto it = hilti::rt::filesystem::recursive_directory_iterator(trimmed_dir, ec);
+        if ( ! ec ) {
+            while ( it != hilti::rt::filesystem::recursive_directory_iterator() ) {
+                if ( it->is_regular_file() && it->path().extension() == ".hlto" )
+                    loadModule(it->path());
+
+                if ( it.increment(ec); ec ) {
+                    ZEEK_DEBUG(hilti::rt::fmt("Error iterating over %s, skipping any remaining files: %s", trimmed_dir,
```
awelzel, line 791: Is that and below worth a warning or warn once (when errno != EEXIST)? I can picture a debugging session due to non-loaded modules until someone pulls out `strace` to figure out there are actually permission errors.

bbannier: Good point, promoted that one and the one on skipping iteration to warnings.

src/compiler/driver.cc
```diff
@@ -125,18 +125,29 @@ Driver::~Driver() {}
...
     if ( ! relative_to.empty() && file.is_relative() ) {
-        if ( auto p = relative_to / file; hilti::rt::filesystem::exists(p) )
+        auto p = relative_to / file;
+        if ( auto exists = hilti::rt::filesystem::exists(p, ec); ! ec && exists )
             file = p;
+        if ( ec ) {
+            return hilti::rt::result::Error(
+                hilti::util::fmt("error computing path of %s relative to %s: %s", file, relative_to, ec.message()));
+        }
     }
```
rsmmr, line 137: Here and also re/ the other changes below: I find this hard to read; suggest to generally change the pattern (1) execute operation, (2) check for error and potentially abort/return; and (3) proceed normally. So like:

```
auto exists = hilti::rt::filesystem::exists(p, ec);
if ( ec )
    return Error(...)

file = p;
...
```

rsmmr (dismissed):
I still find these combined if-conditions non-intuitive, but maybe that's just the weird boolean semantics of `std::error_code`. Fine with me to go with this.

awelzel (approved):
```
$ ZEEK_SPICY_MODULE_PATH=/tmp/modules/ zeek -N                                                   
[libhilti] Warning: Error iterating over /tmp/modules/, skipping any remaining files: Permission denied
```
Seems to work, crash is gone. The `[libhilti]` part may be confusing to users, but there's now some message where previously we'd crash without explanation, so :ship:

bbannier:
> I still find these combined if-conditions non-intuitive, but maybe that's just the weird boolean semantics of `std::error_code`. Fine with me to go with this.

I never pushed the fixes for that, updated the PR now.

PR #184, opened 2023-02-13
Remove unneeded(?) symlink.

### LINKED ISSUE #123: Remove symlink in build tree when built as builtin plugin
If built as a builtin plugin we currently create a link in into the build/source directory and it is not clear anymore exactly why that link was needed. We added a workaround for binary packages when fixing #119, but should probably just get rid of the link.

rsmmr:
It's unclear what this is for. In local testing, I don't seem to need
it. Let's see if anything breaks ...

Closes #123.


rsmmr:
This seems fine. https://github.com/zeek/spicy/pull/1380 also ran it from the Spicy side, likewise with no issues.

bbannier (approved):

PR #186, opened 2023-02-14
Remove unnecessary copying when iterating collections.

bbannier:


bbannier:
Also backported to `release/1.3`.

PR #190, opened 2023-02-21
Include zeek-version.h if zeek-config.h does not define ZEEK_VERSION_…

awelzel:
…NUMBER

This is for zeek/zeek#2776 where zeek-config.h will not provide
ZEEK_VERSION and ZEEK_VERSION_NUMBER anymore when building an
internal plugin.

### DIFF
--- a/include/plugin/zeek-compat.h
+++ b/include/plugin/zeek-compat.h
@@ -8,7 +8,13 @@
 
 #include <zeek/zeek-config.h>
 
-#ifdef ZEEK_VERSION_NUMBER
+// When building as internal plugin, the newest Zeek version does not define
+// ZEEK_VERSION_NUMBER through zeek-config.h, it'll be in the new
+// zeek/zeek-version.h header.
+#if __has_include(<zeek/zeek-version.h>)
+#include <zeek/zeek-version.h>
+#endif
+
 #if ZEEK_SPICY_VERSION_NUMBER != ZEEK_VERSION_NUMBER
 #define STR(x) __STR(x)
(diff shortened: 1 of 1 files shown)


awelzel:
@bbannier / @rsmmr - would it be okay to merge this? I'd then go ahead with zeek/zeek#2776.

If there's something about the CI jobs I can do, let me know, but looks quite unrelated.

bbannier:
> @bbannier / @rsmmr - would it be okay to merge this? I'd then go ahead with [zeek/zeek#2776](https://github.com/zeek/zeek/issues/2776).
> 
> If there's something about the CI jobs I can do, let me know, but looks quite unrelated.

I'd really like to see this pass before merging to make sure it also works with older versions.

The CI job is failing since the build for the Spicy deb file is still in progress ATM. Once that finishes the jobs here should be able to make progress after retriggering.

include/plugin/zeek-compat.h
```diff
@@ -8,7 +8,11 @@
 
 #include <zeek/zeek-config.h>
 
-#ifdef ZEEK_VERSION_NUMBER
+//// Include zeek/zeek-version.h when building internally with latest Zeek.
+#ifndef ZEEK_VERSION_NUMBER
```
bbannier, line 12: Let's instead directly check for the header,
```suggestion
#if __has_include("<zeek/zeek-version.h>")
```

bbannier (approved):
Looks good to me, will merge as is under the assumption that zeek/zeek#2776 will get merged in its current form.

PR #192, opened 2023-02-27
Adjust to upstream CMake changes

Neverlord:
Companion PR for https://github.com/zeek/zeek/pull/2799.


rsmmr:
I believe this is no longer relevant with the recent move into Zeek property, closing.

PR #201, opened 2023-03-08
Make most recently opened file available via `$file` in events.

### LINKED ISSUE #200: File Analysis Framework Filename Function
There is a filename listed in files.log (https://docs.zeek.org/en/master/scripts/base/frameworks/files/main.zeek.html#type-Files::Info) but there isn't a way to add a file name via Spicy with the functions available here: https://docs.zeek.org/projects/spicy/en/latest/zeek.html?highlight=file%20analysis%20framework#controlling-zeek-from-spicy

It would be nice to have a `set_file_name` function available in Spicy that will allow population of this field.  

Thanks!

bbannier:
Closes #200.


rsmmr (approved):

tests/zeek/file-analysis-data-in.zeek
```diff
@@ -42,4 +48,15 @@ protocol analyzer spicy::SSH over TCP:
     port 22/tcp,
     replaces SSH;
 
+on SSH::Banner::software -> event have_filename($file, self.file_name);
```
bbannier, line 51: We will need a doc update in `spicy/spicy/doc/zeek.rst` for that: remove the caveat that `$file` is only available in file analyzers.

src/plugin/runtime-support.cc
```diff
@@ -230,14 +230,42 @@ void rt::debug(const Cookie& cookie, const std::string& msg) {
...
     auto cookie = static_cast<Cookie*>(hilti::rt::context::cookie());
     assert(cookie);
 
     if ( auto x = std::get_if<cookie::FileAnalyzer>(cookie) )
         return x->analyzer->GetFile()->ToVal();
-    else
-        throw ValueUnavailable("$file not available", location);
+    else if ( auto* fstate = _file_state(cookie, {}) )
+        if ( auto* f = ::zeek::file_mgr->LookupFile(fstate->fid) )
+            return f->ToVal();
```
rsmmr, line 266: For readability: 

```suggestion
    else if ( auto* fstate = _file_state(cookie, {}) ) {
        if ( auto* f = ::zeek::file_mgr->LookupFile(fstate->fid) )
            return f->ToVal();
    }
```

rsmmr (approved):

PR #203, opened 2023-03-09
Extend `protocol_*` functions so users can pick child analyzer

### LINKED ISSUE #198: protocol_* should have semantics similar to file_* APIs
Currently `protocol_begin` does not return an identifier unlike `file_begin`.  This usually would work fine except in cases where reassembly is needed before the other protocol analyzer can consume the data. The scenario is akin to writing to the previously opened file.

bbannier:
Closes #198.


include/plugin/runtime-support.h
```diff
@@ -294,17 +307,19 @@ void protocol_data_in(const hilti::rt::Bool& is_orig, const hilti::rt::Bytes& da
...
 
 /**
  * Signals EOD to all previously instantiated Zeek-side child protocol
  * analyzers and removes them.
  *
  * @param analyzer ID of analyzer previously instantiated through `protocol_begin()`.
+ * @param h optional handle to the child analyzer to terminate.
  */
-void protocol_end();
+void protocol_end(const std::optional<ProtocolHandle>& h);
```
bbannier, line 322: In addition to wanting to make the handle type opaque I am using a dedicated type of `ProtocolHandle` instead of a `std::optional<integer::safe<uint64_t>>` here since otherwise I run into issues in generated code where we have issue coercing `Null` to the optional safe integer (two conversions). This is similar to the following code in Spicy:

```cpp
std::optional<integer::safe<uint64_t>> foo(std::optional<integer::safe<uint64_t>> s) { return s; }
foo(hilti::rt::Null());
```

src/plugin/runtime-support.cc
```diff
@@ -427,49 +429,86 @@ void rt::protocol_begin(const std::optional<std::string>& analyzer) {
...
+    if ( h ) {
+        auto* child = c->analyzer->FindChild(h->id);
+        if ( ! child )
+            throw ValueUnavailable(hilti::rt::fmt("unknown child analyzer %s", *h));
+
+        child->NextUndelivered(offset, len, is_orig);
+    }
+
+    else
+        c->analyzer->ForwardUndelivered(offset, len, is_orig);
```
bbannier, line 483: I am surprised that we didn't see this causing issues before. Is this a correct change @rsmmr?

spicy/zeek.spicy
```diff
@@ -43,36 +43,45 @@ public function flip_roles() : void &cxxname="spicy::zeek::rt::flip_roles";
...
 ## Adds a Zeek-side child protocol analyzer to the current connection.
 ##
 ## analyzer: type of analyzer to instantiate, specified through its Zeek-side
 ## name (similar to what Zeek's signature action `enable` takes); if not
 ## specified, Zeek will perform its usual dynamic protocol detection to figure
 ## out how to parse the data (the latter will work only for TCP protocols, though.)
-public function protocol_begin(analyzer: optional<string> = Null) : void &cxxname="spicy::zeek::rt::protocol_begin";
+##
+## Returns identifier of the child analyzer.
+public function protocol_begin(analyzer: optional<string> = Null) : optional<ProtocolHandle> &cxxname="spicy::zeek::rt::protocol_begin";
```
bbannier, line 57: Ideally we'd have integration tests for this explicit management of protocol handles.

src/plugin/runtime-support.cc
```diff
@@ -427,49 +429,86 @@ void rt::protocol_begin(const std::optional<std::string>& analyzer) {
...
-    c->analyzer->ForwardStream(data.size(), reinterpret_cast<const u_char*>(data.data()), is_orig);
+    auto len = data.size();
+    auto* data_ = reinterpret_cast<const u_char*>(data.data());
+
+    if ( h ) {
+        auto* child = c->analyzer->FindChild(h->id);
+        if ( ! child )
+            throw ValueUnavailable(hilti::rt::fmt("unknown child analyzer %s", *h));
+
+        child->DeliverStream(len, data_, is_orig);
```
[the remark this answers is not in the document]
bbannier: Since we have access to the output handler and can trigger the lifecycle management of the child list, it seems we can implement this from the outside.

src/plugin/runtime-support.cc
```diff
@@ -427,49 +429,98 @@ void rt::protocol_begin(const std::optional<std::string>& analyzer) {
 
         if ( ! c->analyzer->AddChildAnalyzer(child) )
             // Same comment as above re/ ignoring the error and memory mgmt.
-            return;
+            return child->GetID();
 
         child->FirstPacket(true, nullptr);
         child->FirstPacket(false, nullptr);
+
+        return {};
```
rsmmr, line 437: Why. not return the ID here?

include/plugin/runtime-support.h
```diff
@@ -270,22 +274,54 @@ void confirm_protocol();
  */
 void reject_protocol(const std::string& reason);
 
+/**
+ * Opaque handle to a protocol analyzer.
+ */
+class ProtocolHandle {
```
rsmmr, line 280: This isn't quite symmetric to how files do it: there we return a string. Maybe we should have originally introduced a `FileHandle` there, but I'm wondering if for symmetry using a string for protocols would be more consistent? But not fully sure what I like more.

include/plugin/runtime-support.h
```diff
@@ -270,22 +274,54 @@ void confirm_protocol();
...
+public:
+    ProtocolHandle() {}
+    ProtocolHandle(uint64_t id) : _id(id) {}
+
+    uint64_t id() const {
+        if ( ! _id )
+            throw ValueUnavailable("uninitialized protocol handle");
```
rsmmr, line 287: This shouldn't really be happening because we're wrapping this into `optional`, right? Then it would at most be an internal error I'd say.

src/plugin/runtime-support.cc
```diff
@@ -427,49 +429,98 @@ void rt::protocol_begin(const std::optional<std::string>& analyzer) {
...
     if ( ! c )
         throw ValueUnavailable("no current connection available");
 
-    c->analyzer->ForwardStream(data.size(), reinterpret_cast<const u_char*>(data.data()), is_orig);
+    auto len = data.size();
+    auto* data_ = reinterpret_cast<const u_char*>(data.data());
+
+    if ( h ) {
+        if ( auto* output_handler = c->analyzer->GetOutputHandler() )
+            output_handler->DeliverStream(len, data_, is_orig);
```
rsmmr, line 455: I  need to re-visit the use of output_handlers, but I'm wondering if it's the right thing to do to send the data through any handler if we pass it on to only one child.

(Same below).

src/plugin/runtime-support.cc
```diff
@@ -427,49 +429,98 @@ void rt::protocol_begin(const std::optional<std::string>& analyzer) {
...
+
+        // Forward empty payload to trigger lifecycle management in this analyzer tree.
+        c->analyzer->ForwardStream(0, data_, is_orig);
+
+        auto* child = c->analyzer->FindChild(h->id());
+        if ( ! child || child->IsFinished() || child->Removing() )
+            throw ValueUnavailable(hilti::rt::fmt("unknown child analyzer %s", *h));
```
rsmmr, line 462: Split into two error cases: "unknown analyzer" and "no longer exists"; might help tracking down issues.

(Same below).

rsmmr (commented):
I'm feeling a bit uneasy putting this low-level logic into the plugin; seems the better place would be the Zeek analyzer API. That said, the plugin will move into Zeek soon, and that makes feel better about it (and it'll be easier to change the Zeek API in lockstep), so I think I'm ok with this for now, minus the comments on semantics.

spicy/zeek.spicy
```diff
@@ -43,36 +43,39 @@ public function flip_roles() : void &cxxname="spicy::zeek::rt::flip_roles";
...
```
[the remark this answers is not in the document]
bbannier: The semantics here would be to not hand out a handle if users are using DPD (not providing a concrete `analyzer` arg in the call). If it makes sense to also hand out a handle to the DPD tree we could always return a value.

src/plugin/runtime-support.cc
```diff
@@ -415,10 +415,12 @@ void rt::protocol_begin(const std::optional<std::string>& analyzer) {
             // need to track what the other side already did. Note that
             // AddChildAnalyzer() will have deleted child already, so nothing
             // for us to clean up here.
-            return;
+            return child->GetID();
```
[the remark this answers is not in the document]
bbannier: I think users do not necessarily see that we added nothing, so I think it could be useful.

src/plugin/runtime-support.cc
bbannier: See above.

include/plugin/runtime-support.h
bbannier: The issue with returning a `string` would be that we would need to parse it whenever we make use of the handle. I believe we should avoid that since it would add unnecessary cost to each call using a handle.

include/plugin/runtime-support.h
bbannier: Since `ProtocolHandle` is a rt type it needs to be default constructible. Such a default-constructed handle has no meaningful ID and we need to reject them.

rsmmr (commented):
Left final comments after another look. Main question left is how to handle an already existing analyzer of the same kind. Everything else looks good to me.

include/plugin/runtime-support.h
```diff
@@ -270,22 +272,54 @@ void confirm_protocol();
...
+std::optional<ProtocolHandle> protocol_begin(const std::optional<std::string>& analyzer);
 
 /**
  * Forwards data to all previously instantiated Zeek-side child protocol
  * analyzers.
  *
  * @param is_orig true to feed data to originator side, false for responder
  * @param data next chunk of stream data for child analyzer to process
+ * @param h optional handle to the child analyzer to stream data into
  */
```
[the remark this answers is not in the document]
bbannier: I added documentation of the return value and its semantics on `protocol_begin` above. This function returns nothing.

rsmmr (approved):

Mohan-Dhawan:
I tried the patch to implement only one instance of the child analyzer (so that I can maintain state information via contexts) and it works. Thanks!

include/plugin/runtime-support.h
```diff
@@ -278,14 +310,28 @@ void reject_protocol(const std::string& reason);
...
 void protocol_begin(const std::optional<std::string>& analyzer);
 
+/**
+ * Gets a handle to the given child analyzer. If the child does not yet exist
+ * it will be created.
+ *
+ * @param analyzer the Zeek-side name of the analyzer to get
```
rsmmr, line 317:
```suggestion
 * Gets a handle to a child analyzer of a given type. If a child of that type does not yet exist
 * it will be created.
 *
 * @param analyzer the Zeek-side name of the analyzer to get (e.g., `HTTP`)
```

include/plugin/runtime-support.h
```diff
@@ -294,18 +340,27 @@ void protocol_data_in(const hilti::rt::Bool& is_orig, const hilti::rt::Bytes& da
...
 void protocol_gap(const hilti::rt::Bool& is_orig, const hilti::rt::integer::safe<uint64_t>& offset,
-                  const hilti::rt::integer::safe<uint64_t>& len);
+                  const hilti::rt::integer::safe<uint64_t>& len, const std::optional<ProtocolHandle>& h = {});
 
 /**
  * Signals EOD to all previously instantiated Zeek-side child protocol
  * analyzers and removes them.
  *
  * @param analyzer ID of analyzer previously instantiated through `protocol_begin()`.
+ * @param h optional handle to the child analyzer to terminate.
```
rsmmr, line 353: There are no parameters anymore.

spicy/zeek.spicy
```diff
@@ -43,37 +43,59 @@ public function flip_roles() : void &cxxname="spicy::zeek::rt::flip_roles";
...
 ## out how to parse the data (the latter will work only for TCP protocols, though.)
 public function protocol_begin(analyzer: optional<string> = Null) : void &cxxname="spicy::zeek::rt::protocol_begin";
 
+## Gets a handle to a Zeek-side child protocol analyzer for the current connection.
+##
+## If no such child exists it will be added; otherwise a handle to the
+## existing child protocol analyzer will be returned.
+##
+## analyzer: type of analyzer to instantiate, specified through its Zeek-side
+## name (similar to what Zeek's signature action `enable` takes).
```
rsmmr, line 67: Likewise document the error semantics here.

spicy/zeek.spicy
```diff
@@ -43,37 +43,59 @@ public function flip_roles() : void &cxxname="spicy::zeek::rt::flip_roles";
...
 ##
+## If the same analyzer was added previously with protocol_handle_get_or_create or
+## protocol_begin with same argument, and not closed with protocol_handle_close
+## or protocol_end, no new analyzer will be added.
+##
 ## analyzer: type of analyzer to instantiate, specified through its Zeek-side
 ## name (similar to what Zeek's signature action `enable` takes); if not
 ## specified, Zeek will perform its usual dynamic protocol detection to figure
 ## out how to parse the data (the latter will work only for TCP protocols, though.)
 public function protocol_begin(analyzer: optional<string> = Null) : void &cxxname="spicy::zeek::rt::protocol_begin";
```
rsmmr, line 59: Document the error semantics here. I see in the code, that errors trigger exceptions.

rsmmr (approved):

PR #204, opened 2023-03-10
Fix arg mixup in call `protocol_gap`.

bbannier:
We seem to have passed arguments in the wrong order which unfortunately compiled successfully due to implicit conversions.


rsmmr (approved):
Argh.

Issue #202, opened 2023-03-09
Allow users to inject additional spicyz options in `spicy_add_analyzer`


bbannier:
If one currently wants to inject additional flags like e..g, `-T` or `-X flow` into the `spicyz` invocation generated from our CMake macro `spicy_add_analyzer` one needs to patch `cmake/ZeekSpicyAnalyzerSupport.cmake` in the installation, e.g., somewhere after https://github.com/zeek/spicy-plugin/blob/dc608c98a6e89af6d45166580831a1f8f12b367e/cmake/ZeekSpicyAnalyzerSupport.cmake#L143-L148

We should instead expose a CMake option which users can modify.

PR #205, opened 2023-03-15
Allow passing spicyz flags via cmake -DSPICYZ_FLAGS=...

simeonmiteff:
This is a fix for #202.

One thing I'm uncertain about is whether `-DSPICYZ_FLAGS` should _override_ the `-d` logic for Debug builds (as it does now) or if `-d` should be appended to whatever flags are provided.


rsmmr:
Thanks for tackling this!

> One thing I'm uncertain about is whether `-DSPICYZ_FLAGS` should _override_ the `-d` logic for Debug builds (as it does now) or if `-d` should be appended to whatever flags are provided.

I'd append them, as that seems the more common semantics for such flags and avoids needing knowledge about what the defaults would be otherwise. 

Also, what do do you think about allowing to set this from the environment as well, similar to what one can do for, e.g., CFLAGS?

simeonmiteff:
> I'd append them, as that seems the more common semantics for such flags and avoids needing knowledge about what the defaults would be otherwise.

Got it, amended commit 👍🏻 

> Also, what do do you think about allowing to set this from the environment as well, similar to what one can do for, e.g., CFLAGS?

Agreed. I've kept the env the same (`SPICYZ_FLAGS`) and allowed specifying both rather than having one override the other. This is also included that in the amended commit.

Is this worth surfacing explicitly in the documentation (perhaps under [7.4. Compiling Analyzers](https://docs.zeek.org/projects/spicy/en/latest/zeek.html#compiling-analyzers))?

cmake/ZeekSpicyAnalyzerSupport.cmake
```diff
@@ -140,12 +140,20 @@ if (SPICYZ)
...
     endif ()
 
+    # Allow passing SPICYZ_FLAGS via environment variable
+    set(SPICYZ_FLAGS "${SPICYZ_FLAGS};$ENV{SPICYZ_FLAGS}")
+
+    if ("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
+        set(SPICYZ_FLAGS "${SPICYZ_FLAGS};-d")
+    endif ()
+
+    string(REPLACE " " ";" SPICYZ_FLAGS "${SPICYZ_FLAGS}")
```
rsmmr, line 155: Just a nit: I think I'd apply this earlier just to the value of `$ENV{}`, so that `SPICYZ_FLAGS` is valid list all the time.

rsmmr:
> Agreed. I've kept the env the same (`SPICYZ_FLAGS`) and allowed specifying both rather than having one override the other. 

Cool, thanks. I left one more comment, otherwise just take it out of draft and I'll merge.

> Is this worth surfacing explicitly in the documentation (perhaps under [7.4. Compiling Analyzers](https://docs.zeek.org/projects/spicy/en/latest/zeek.html#compiling-analyzers))?

Definitely worth documenting, just wondering where the best place is. We currently don't really talk about the CMake setup anywhere, and not sure how helpful it would be there in 7.4. I think the closest is the part on Zeek packages, https://docs.zeek.org/projects/spicy/en/latest/tutorial/index.html#creating-a-zeek-package. Or maybe we really need a section on the CMake pieces. What do you think?

cmake/ZeekSpicyAnalyzerSupport.cmake
simeonmiteff: Ah, hmmm 🤔... so, when `SPICYZ_FLAGS` is set via `cmake -D` then it starts life as a non-valid list (as `$ENV{SPICYZ_FLAGS}` would), so I figured: wait for all the concatenations before doing the `string(REPLACE, ...)`.

Do you know if there is another path by which `SPICYZ_FLAGS` might be set as a valid list instead?

Full disclosure: I'm still a CMake noob, so I found this list business somewhat puzzling (initially spaces were being escaped in the resulting `spicyz` command line).

rsmmr: Ok, I reworked this a little bit mainly for clarity: https://github.com/zeek/spicy-plugin/compare/main...topic/robin/cmake-user-spicyz-flags.

bbannier:
Sorry for arriving to the discussion late.

I think since we want to preserve the implicit flags (`-d` for `Debug` builds), and allow supplying additional flags via CMake command line flags (`-DSPICY_FLAGS`) as well as through an environment variable `SPICY_FLAGS` we should only allow only of them at a time (probably: explicit command line flag overrides env var).

It would also be great if we could expose the current CMake variable as an [`option`](https://cmake.org/cmake/help/latest/command/option.html) (default empty) so it is more easily discoverable via e.g., `ccmake` or command line completion. In that case the arg to `spicyz` would be a new variable, containing optionally the implicit `-d` followed by whatever the user specified.

rsmmr:
> we should only allow only of them at a time (probably: explicit command line flag overrides env var).

I actually prefer adding them up. I don't think it will matter much either way in the end, so I'll leave it at that.

> It would also be great if we could expose the current CMake variable as an [`option`]

Ok, I made that change in https://github.com/zeek/spicy-plugin/pull/208.

PR #211, opened 2023-04-05
Fix spelling in sources and file names

bbannier:


rsmmr (commented):
What happens if the pre-commit reports a false positive?

bbannier:
> What happens if the pre-commit reports a false positive?

I would suggest that we work around that on a case-by-case basis.

From using the same hook over in zeek/spicy for more than a year we have mostly seen this in two scenarios:

- It might mistake combined flags in shell code snippets as misspelled text, e.g., it might flag the flags `-ot` as a misspelling of `-or`. One fix could be to use `-o -t`.
- When running on Git commit messages it might identify a Git commit hash in a comment in the commit message as a misspelled identifier. I usually work around this by skipping the lint with `SKIP=gitlint git commit ...` or by skipping hooks completely with `git commit -n ...`.

If none of these help we could allow certain identifiers with a config file (see e.g., [how we do that in Spicy](https://github.com/zeek/spicy/blob/main/.typos.toml)), or spellchecking for whole files with a [pre-commit exclusion](https://pre-commit.com/#config-exclude).

rsmmr:
> see e.g., [how we do that in Spicy]

Ah, I forgot about that whitelist capability.

rsmmr (approved):

PR #214, opened 2023-04-20
More performance improvements

rsmmr:


src/plugin/file-analyzer.cc
```diff
@@ -67,7 +67,7 @@ bool FileAnalyzer::EndOfFile() {
 
 bool FileAnalyzer::Process(int len, const u_char* data) {
     if ( ! _state.hasParser() && ! _state.isSkipping() ) {
-        auto parser = OurPlugin->parserForFileAnalyzer(_state.cookie().analyzer->Tag());
+        auto parser = OurPlugin->parserForFileAnalyzer(_state.file().analyzer->Tag());
         ;
```
bbannier, line 71: Nit: stray `;`.

include/plugin/cookie.h
```diff
@@ -112,7 +112,121 @@ struct PacketAnalyzer {
...
+
+        new (&data) Data(tag(), std::move(other.data));
+        return *this;
+    }
+
+    // Cache of values that can be expensive to compute.
+    struct {
+        ::zeek::ValPtr conn = nullptr;
+        ::zeek::ValPtr is_orig = nullptr;
+        bool confirmed = false;
```
bbannier, line 148: Could you add comments to the members that each of them is only meaningful for `protocol`?

src/plugin/plugin.cc
```diff
@@ -60,7 +61,9 @@ plugin::Zeek_Spicy::Plugin::Plugin() {
 
 void ::spicy::zeek::debug::do_log(const std::string& msg) {
     PLUGIN_DBG_LOG(*plugin::Zeek_Spicy::OurPlugin, "%s", msg.c_str());
-    HILTI_RT_DEBUG("zeek", msg);
+
+    if ( hilti::rt::isInitialized() )
+        HILTI_RT_DEBUG("zeek", msg);
```
bbannier, line 66: Is this required because of the changes over in Spicy? I am asking since in that case it would be much nicer if `HILTI_RT_DEBUG` handled checking its requirements internally instead of every user needing to think about it.

include/plugin/runtime-support.h
```diff
@@ -710,6 +709,95 @@ inline ::zeek::ValPtr to_val(const hilti::rt::Set<T>& s, ::zeek::TypePtr target)
...
+template<template<typename...> typename U, typename... Ts>
+struct is_instance_impl<U<Ts...>, U> : public std::true_type {};
+} // namespace
+
+template<typename T, template<typename...> typename U>
+using is_instance = is_instance_impl<std::remove_cv_t<T>, U>;
+
+
+template<typename T>
+inline void set_record_field(::zeek::RecordVal* rval, ::zeek::IntrusivePtr<::zeek::RecordType>& rtype, int idx,
```
bbannier, line 726: Is `rtype` mutable due to const-correctness issues in `IntrusivePtr`/`RecordType`?

include/plugin/runtime-support.h
```diff
@@ -710,6 +709,95 @@ inline ::zeek::ValPtr to_val(const hilti::rt::Set<T>& s, ::zeek::TypePtr target)
     return zv;
 }
 
+
+namespace {
+template<typename, template<typename...> typename>
+struct is_instance_impl : public std::false_type {};
```
bbannier, line 715: Nit: `public` is implicit.

Same for specialization below.

include/plugin/runtime-support.h
```diff
@@ -710,6 +709,95 @@ inline ::zeek::ValPtr to_val(const hilti::rt::Set<T>& s, ::zeek::TypePtr target)
...
+
+template<typename T, template<typename...> typename U>
+using is_instance = is_instance_impl<std::remove_cv_t<T>, U>;
+
+
+template<typename T>
+inline void set_record_field(::zeek::RecordVal* rval, ::zeek::IntrusivePtr<::zeek::RecordType>& rtype, int idx,
+                             const T& x) {
+    using NoConversionNeeded =
+        std::integral_constant<bool, std::is_same<T, int8_t>::value || std::is_same<T, int16_t>::value ||
```
bbannier, line 729: Nit: this function could be less busy if we'd use `std::is_same_v`.

include/plugin/runtime-support.h
```diff
@@ -710,6 +709,95 @@ inline ::zeek::ValPtr to_val(const hilti::rt::Set<T>& s, ::zeek::TypePtr target)
...
+        rval->Assign(idx, static_cast<int64_t>(x.Ref()));
+#else
+        rval->Assign(idx, static_cast<int>(x.Ref()));
+#endif
+    else if constexpr ( IsUnsignedInteger::value )
+        rval->Assign(idx, static_cast<uint64_t>(x.Ref()));
+    else if constexpr ( std::is_same<T, hilti::rt::Bytes>::value )
+        rval->Assign(idx, x.str());
+    else if constexpr ( std::is_same<T, hilti::rt::Bytes>::value )
+        rval->Assign(idx, x.str());
```
bbannier, line 761: This duplicates the branch above. Should this have been something else or is this just left-over copy-n-paste?

bbannier (approved):
LGTM, mostly left smaller nits.

src/plugin/plugin.cc
rsmmr: Yes and no: The changes triggered it, but it was ill-defined to begin with: generally HILTI runtime functionality isn't usable before the library has been initialized. It just happened to work so far. That also addresses the 2nd part: I think it's reasonable for the macro to assume that the runtime has been initialized already (just like pretty much any other runtime functionality does).

include/plugin/runtime-support.h
rsmmr: Oversight.

include/plugin/runtime-support.h
rsmmr: Left-over indeed.

